PR #67, opened 2020-07-09
fixed address parsing crash for incorrect value

gagarin55:


sigma-tree/src/chain/address.rs
```diff
@@ -280,5 +280,11 @@ mod tests {
             let decoded_addr = encoder.parse_address_from_str(&encoded_addr).unwrap();
             prop_assert_eq![decoded_addr, v];
         }
+
+        #[test]
+        fn doesnt_crash(s in "\\w+") {
```
greenhat, line 285:
```suggestion
        fn doesnt_crash_on_invalid_input(s in "\\w+") {
```

sigma-tree/src/chain/address.rs
```diff
@@ -219,7 +219,7 @@ impl AddressEncoder {
     /// parse address from Base58 encoded string
     pub fn parse_address_from_str(&self, str: &str) -> Result<Address, AddressEncoderError> {
         let bytes = bs58::decode(str).into_vec()?;
-        if bytes.is_empty() {
+        if bytes.len() < AddressEncoder::CHECKSUM_LENGTH {
```
greenhat, line 222: Good catch! Thanks! 
This check does eliminate panicking down the line but I feel it would be hard to reason about for a newcomer without drilling down the code. I'm thinking of introducing a `const MIN_ADDRESS_LENGTH: usize = AddressEncoder::CHECKSUM_LENGTH + 2 // 1 for header byte + 1 for min content bytes length` and using it in this check like `bytes.len() < AddressEncoder::MIN_ADDRESS_LENGTH`. What do you think?

greenhat: Great! Thanks!

PR #69, opened 2020-07-22
make some Digest32, ProverResult and ContextExtension public types 

gagarin55:
Having these types public is critical for `ErgoBoxCandidate` construction


sigma-tree/src/chain/digest32.rs
```diff
@@ -14,6 +14,7 @@ use std::convert::TryInto;
 use std::io;
 use thiserror::Error;
 
+///
```
greenhat, line 17:
```suggestion
/// 32 byte array used in box, transaction ids (hash)
```

PR #81, opened 2020-08-17
Finish EcPoint::exponentiate implementation.

jasondavies:
I think it's already finished!

Scalar is already modulo the group order, so no special check for negative exponent is necessary.


greenhat:
Also, there is some progress with Scalar arithmetic ops from new RustCrypto libs. Check out https://github.com/ergoplatform/sigma-rust/pull/80

PR #133, opened 2020-11-05
First draft of extended wasm addresses

SebastienGllmt:
For Yoroi, we use the following ergo-ts functions

```
fromBytes
fromBase58
getType
fromPk
fromErgoTree
addr
```

Currently the only thing exposed by sigma-rust is

```
declare export class Address {
    static from_testnet_str(s: string): Address;
}
```

This PR is meant to add bindings to expose the missing functionality


greenhat:
Continues in #136

PR #144, opened 2020-11-16
Add functionality needed for Yoroi

SebastienGllmt:
This PR adds all the functionality we needed for Yoroi. It's kind of a large PR so I need to rebase it and probably split it up into smaller chunks. Just putting this PR up for visibility though.


greenhat:
I’ve glanced through the diff and it does not look that big to me. I don’t think splitting is necessary.

ergo-lib/src/wallet/tx_builder.rs
```diff
@@ -131,15 +162,26 @@ impl<S: ErgoBoxAssets + ErgoBoxId + Clone> TxBuilder<S> {
...
             .map(|b| {
-                ErgoBoxCandidateBuilder::new(
+                let mut candidate = ErgoBoxCandidateBuilder::new(
                     b.value,
                     change_address_ergo_tree.clone(),
                     self.current_height,
-                )
-                .build()
+                );
+                for token in &b.tokens() {
```
SebastienGllmt, line 170: Note: this is a change in the behavior for how TxBuilder handles change. It used not include tokens inside the change address, which means that transactions that didn't explicitly send input tokens to an output failed.

ergo-lib/src/wallet/tx_builder.rs
```diff
@@ -131,15 +162,26 @@ impl<S: ErgoBoxAssets + ErgoBoxId + Clone> TxBuilder<S> {
...
+                    candidate.add_token(token.clone());
+                }
+                candidate.build()
             })
             .collect();
         output_candidates.append(&mut change_boxes?);
+
+        // Ergo transactions need at least one output
+        // but that single output could just be the change (ex: if you want to send all ERG to a single address)
+        if output_candidates.is_empty() {
```
SebastienGllmt, line 180: Note: this fixes the issue where you couldn't use the TxBuilder to create a transaction with only change addresses (no explicit outputs). We use this kind of transaction in Yoroi for send-all transactions

ergo-lib/src/chain/ergo_box/box_value.rs
```diff
@@ -24,16 +24,19 @@ impl BoxValue {
...
     /// Absolute minimal value, calculated from smallest possible box size and original value per byte requirement
     pub const MIN: BoxValue = BoxValue(BoxValue::MIN_RAW);
 
     /// Recommended (safe) minimal box value to use in case box size estimation is unavailable.
-    /// Allows box size upto 2777 bytes with current min box value per byte of 360 nanoERGs
+    /// Allows box size up to 2777 bytes with current min box value per byte of 360 nanoERGs
     pub const SAFE_USER_MIN: BoxValue = BoxValue(1000000);
 
+    /// Number of units inside one ERGO (i.e. one ERG using nano ERG representation)
+    pub const UNITS_PER_ERGO: u32 = 1000000000;
```
SebastienGllmt, line 38: I decided to put this in BoxValue, but if you prefer it being elsewhere that is fine too

ergo-lib/src/chain/address.rs
```diff
@@ -303,18 +368,25 @@ impl AddressEncoder {
 
     /// encode address as Base58 encoded string
     pub fn address_to_str(&self, address: &Address) -> String {
-        AddressEncoder::encode_address(self.network_prefix, &address)
+        AddressEncoder::encode_address_as_string(self.network_prefix, &address)
     }
 
     /// encode address as Base58 encoded string
-    pub fn encode_address(network_prefix: NetworkPrefix, address: &Address) -> String {
+    pub fn encode_address_as_bytes(network_prefix: NetworkPrefix, address: &Address) -> Vec<u8> {
```
SebastienGllmt, line 375: Just an FYI this is a breaking change in the API. I did this because in Yoroi we prefer handling addresses as byte arrays everywhere and this refactoring allows you to handle both bytes and base58

ergo-lib/src/chain/address.rs
```diff
@@ -265,24 +310,44 @@ impl AddressEncoder {
...
+
+    /// parse address from Base58 encoded string
+    pub fn unchecked_parse_address_from_str(str: &str) -> Result<Address, AddressEncoderError> {
+        let bytes = bs58::decode(str).into_vec()?;
+        AddressEncoder::unchecked_parse_address_from_bytes(&bytes)
     }
 
     /// parse address from Base58 encoded string
-    fn try_parse_address(bytes: &[u8]) -> Result<Address, AddressEncoderError> {
+    pub fn unchecked_parse_address_from_bytes(bytes: &[u8]) -> Result<Address, AddressEncoderError> {
```
SebastienGllmt, line 347: Note: these "unchecked" functions are all to power the NetworkAddress struct which keeps track of both the network&address in the same struct. Since the NetworkAddress class explicitly takes in a network as a parameter, it's easier if we skip AddresssEncoder network check

bindings/ergo-lib-wasm/src/ergo_box/box_builder.rs
```diff
@@ -112,8 +112,9 @@ impl ErgoBoxCandidateBuilder {
     }
 
     /// Build the box candidate
-    pub fn build(self) -> Result<ErgoBoxCandidate, JsValue> {
+    pub fn build(&self) -> Result<ErgoBoxCandidate, JsValue> {
```
SebastienGllmt, line 115: FYI these used to destroy the builder, but after this PR the old object is kept in memory until you `free` it explicitly. This is useful for us in Yoroi because the way the wallet works is that the transaction confirmation page actually populates the UI directly from a TxBuilder. It called the `build()` once to populate the UI (to show the confirmation page), then called it again later to actually sign the transaction. It's much easier to keep the builder in memory than keeping the `UnsignedTransaction` in memory since the builder retains more information about how the tx was built (ex: which addresses were change, etc.)

bindings/ergo-lib-wasm/src/ergo_box.rs
```diff
@@ -163,10 +198,16 @@ impl BoxValue {
...
-    /// Create from u32 with bounds check
-    pub fn from_u32(v: u32) -> Result<BoxValue, JsValue> {
+    /// Number of units inside one ERGO (i.e. one ERG using nano ERG representation)
+    #[allow(non_snake_case)]
+    pub fn UNITS_PER_ERGO() -> I64 {
+        (chain::ergo_box::BoxValue::UNITS_PER_ERGO as i64).into()
+    }
+
+    /// Create from i64 with bounds check
+    pub fn from_i64(v: &I64) -> Result<BoxValue, JsValue> {
```
SebastienGllmt, line 208: I got rid of the `from_u32` here as per our discussion that it was better to replace it with `from_i64`

bindings/ergo-lib-wasm/src/ergo_box.rs
```diff
@@ -131,10 +166,10 @@ impl ErgoBox {
             .map(Constant::from)
     }
 
-    // JSON representation
-    // pub fn to_json(&self) -> Result<JsValue, JsValue> {
-    //     JsValue::from_serde(&self.0).map_err(|e| JsValue::from_str(&format!("{}", e)))
-    // }
+    /// JSON representation
+    pub fn to_json(&self) -> Result<JsValue, JsValue> {
+        JsValue::from_serde(&self.0.clone()).map_err(|e| JsValue::from_str(&format!("{}", e)))
```
SebastienGllmt, line 171: FYI I added `.clone()` to all `to_json` functions because otherwise the `to_json` would free to underlying WASM object which would cause a crash if you used the object again after calling `to_json` on it (which wasn't obvious)

bindings/ergo-lib-wasm/src/ergo_box.rs
```diff
@@ -98,12 +117,13 @@ impl ErgoBox {
         contract: &Contract,
         tx_id: &TxId,
         index: u16,
+        tokens: &Tokens,
```
SebastienGllmt, line 120: recall: we needed this so that we can build a transaction in WASM and use WASM to calculate the `boxId`. The `from_json` function requires the boxId as a parameter (which there was no way of knowing if for example you're creating boxes for mock data)

bindings/ergo-lib-wasm/src/address.rs
```diff
@@ -179,3 +209,54 @@ impl From<chain::address::Address> for Address {
         Address(a)
     }
 }
+
+/// Combination of an Address with a network
+/// These two combined together form a base58 encoding
+#[wasm_bindgen]
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct NetworkAddress(chain::address::NetworkAddress);
```
SebastienGllmt, line 217: I created a `NetworkAddress` since it's easier to work with in Yoroi if we can just keep track of which network the address was for everywhere since Yoroi supports both testnet & mainnest all from the same extension.

ergo-lib/src/chain/address.rs
```diff
@@ -265,24 +310,44 @@ impl AddressEncoder {
     /// parse address from Base58 encoded string
     pub fn parse_address_from_str(&self, str: &str) -> Result<Address, AddressEncoderError> {
         let bytes = bs58::decode(str).into_vec()?;
-        if bytes.len() < AddressEncoder::MIN_ADDRESS_LENGTH {
+        if bytes.len() < 1 {
```
greenhat, line 313: What is the reason for this change?

ergo-lib/src/chain/address.rs
```diff
@@ -265,24 +310,44 @@ impl AddressEncoder {
...
+    /// parse network+address from Base58 encoded string
+    pub fn unchecked_parse_network_address_from_str(str: &str) -> Result<NetworkAddress, AddressEncoderError> {
         let bytes = bs58::decode(str).into_vec()?;
-        if bytes.len() < AddressEncoder::MIN_ADDRESS_LENGTH {
+        AddressEncoder::unchecked_parse_network_address_from_bytes(&bytes)
+    }
+
+    /// parse network+address from Base58 encoded string
+    pub fn unchecked_parse_network_address_from_bytes(bytes: &[u8]) -> Result<NetworkAddress, AddressEncoderError> {
+        if bytes.len() < 1 {
```
greenhat, line 328: Why not use `AddressEncoder::MIN_ADDRESS_LENGTH` here?

ergo-lib/src/wallet/tx_builder.rs
greenhat: Oops, my bad. Forgot to put the tokens. Good catch! Thanks for the fix!

greenhat (changes requested):
Thank you, it looks good! I appreciate your notes! See some minor comments. Also, please check red linter and formatter CI jobs.

ergo-lib/src/chain/address.rs
SebastienGllmt: because the check for `AddressEncoder::MIN_ADDRESS_LENGTH` was moved to `unchecked_parse_address_from_bytes`. I still need a check of `>= 1` in this function though because of the line
```
let network_prefix = (bytes[0] & 0xF0).try_into()?;
```

I could check for `AddressEncoder::MIN_ADDRESS_LENGTH` in both places, but since this function only needs a header byte, I made the requirement be just `>= 1`

ergo-lib/src/chain/address.rs
SebastienGllmt: same as above comment

ergo-lib/src/chain/address.rs
greenhat: Got it. I don't want to nitpick, but I find that the new check has some drawbacks comparing to using the constant:
-  "magic" number is harder to reason about;
-  with a 1-byte array the function will fail later anyway.

greenhat (approved):
Thank you!

PR #176, opened 2021-01-20
Optional boxid for json conversion

SebastienGllmt:
# Goal
Using `ErgoBox.from_json` with dummy data for testing purposes were we may not have the boxId calculated.

# Problem

Currently not passing a boxId (or passing in a dummy value) throws an error (`Box id parsed from JSON differs from calculated from box serialized bytes`)

# Solution
Use the calculated value for the field instead of throwing an error if the boxId field is missing


greenhat (approved):
Thanks!

PR #226, opened 2021-04-01
Sigma protocol for AND and OR conjectures

### LINKED ISSUE #194: Sigma protocol implementation for And, Or

greenhat:
Close #194 

- [x] `SigmaAnd` evaluation;
- [x] `SigmaOr` evaluation;
- [x] Sigma protocol implementation for AND conjectures;
- [x] Sigma protocol implementation for OR conjectures;
- [x] verifier tests with test vectors confirmed in sigmastate;
- [x] update changelog, readme;

This PR adds an evaluation for AND, OR conjectures in sigma protocol.

The evaluation part is in `eval/sigma_*.rs`.

The entry point for prover implementation is in `prover.rs` in `Prover::prove()`.
where for every step conjecture support was added. 

Although data types for distributed signatures were added (see `prover/hint.rs`) there is no implementation in the prover.

For sigma trees, traversal and node rewrite (similar to kiama's `everywheretd` strategy) `rewrite` function was implemented in `proof_tree.rs`/

The entry point for verifier implementation is in `verifier.rs` in `Verifier::verify()`.

Test vectors verifying the prover and verifier against the reference Scala implementation are in `verifier.rs` in `mod tests` along with property-based roundtrip prover->verifier tests. Corresponding test vectors in Scala version are in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/724


ergotree-interpreter/src/sigma_protocol/fiat_shamir.rs
```diff
@@ -17,19 +17,35 @@ use thiserror::Error;
...
+use super::GROUP_SIZE;
+use super::SOUNDNESS_BYTES;
+
 /// Hash type for Fiat-Shamir hash function (24-bytes)
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub struct FiatShamirHash(pub Box<[u8; SOUNDNESS_BYTES]>);
 
+impl FiatShamirHash {
+    pub fn secure_random() -> Self {
```
kushti, line 30: Why this method is in FiatShamirHash? Should be in something like CryptoUtils

kushti (approved):

ergotree-interpreter/src/sigma_protocol/fiat_shamir.rs
greenhat: Thanks for catching this! I moved a "real" random bytes generator `secure_random_bytes` function to the new `crypto_utils` module. This is just a "wrapper" for `secure_random_bytes` to ensure the size of the generated byte array is correct. It produces an instance of this type (`FiatShamirHash`) so think of it as a constructor for this type.

PR #227, opened 2021-04-06
Add AST node and evaluation for DecodePoint

Shimuuar:
This is first PR for parsing ergo script. In the end we'll need to be able to parse all of ergoscript for ergvein wallet and DecodePoint is as good starting point as any

I wasn't able to find where DecodePoint is evaluated in scala code so this is only my best guess on how it should be evaluated.

P.S. I noticed that there're quite a few of AST nodes that are just wrappers over `Box<Expr>`. At least `And`, `Or`, `DecodePoint` and probably others. Wouldn't it be better to implement them using single struct parametrized by phantom type parameter to distinguish different nodes? Something along the lines of:

```rust
struct UnaryNode<T> {
    input: Box<Expr>
}
```

### DIFF
--- a/ergotree-interpreter/src/eval/decode_point.rs
+++ b/ergotree-interpreter/src/eval/decode_point.rs
@@ -1,0 +1,45 @@
+use ergotree_ir::mir::decode_point::DecodePoint;
+use ergotree_ir::mir::value::Value;
+
+use crate::eval::env::Env;
+use crate::eval::EvalContext;
+use crate::eval::EvalError;
+use crate::eval::EvalError::Misc;
+use crate::eval::Evaluable;
+use ergotree_ir::mir::constant::TryExtractInto;
+use ergotree_ir::serialization::SigmaSerializable;
+use ergotree_ir::sigma_protocol::dlog_group::EcPoint;
+
+impl Evaluable for DecodePoint {
+    fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let point_bytes = self.input.eval(env, ctx)?.try_extract_into::<Vec<u8>>()?;
+        let point: EcPoint = SigmaSerializable::sigma_parse_bytes(point_bytes).map_err(|_| {
+            Misc(String::from(
+                "DecodePoint: Failed to parse EC point from bytes",
+            ))
+        })?;
+        Ok(point.into())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::eval::tests::eval_out_wo_ctx;
+
+    use super::*;
+
+    use ergotree_ir::mir::expr::Expr;
+    use ergotree_ir::serialization::SigmaSerializable;
+    use proptest::prelude::*;
+
+    proptest! {
+
+        #[test]
+        fn eval(ecp in any::<EcPoint>()) {
+            let bytes = ecp.sigma_serialize_bytes();
+            let expr: Expr = DecodePoint {input: Expr::Const(bytes.clone().into()).into()}.into();
+            let res = eval_out_wo_ctx::<EcPoint>(&expr);
+            prop_assert_eq!(res, ecp);
+        }
+    }
+}
--- a/ergotree-ir/src/mir/decode_point.rs
+++ b/ergotree-ir/src/mir/decode_point.rs
@@ -1,0 +1,84 @@
+//! Decode byte array to EC point
+
+use crate::serialization::op_code::OpCode;
+use crate::serialization::sigma_byte_reader::SigmaByteRead;
+use crate::serialization::sigma_byte_writer::SigmaByteWrite;
+use crate::serialization::SerializationError;
+use crate::serialization::SigmaSerializable;
+use crate::types::stype::SType;
+
+use super::expr::Expr;
+
+/// Decode byte array to EC point
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct DecodePoint {
+    /// Byte array to be decoded
+    pub input: Box<Expr>,
+}
+
+impl DecodePoint {
+    pub(crate) const OP_CODE: OpCode = OpCode::DECODE_POINT;
+
+    /// Type
+    pub fn tpe(&self) -> SType {
+        SType::SGroupElement
+    }
+
+    pub(crate) fn op_code(&self) -> OpCode {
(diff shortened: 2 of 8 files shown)


ergotree-interpreter/src/eval/decode_point.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    use ergotree_ir::mir::expr::Expr;
+    use proptest::collection;
+    use proptest::prelude::*;
+    use sigma_test_util::force_any_val;
+
+    proptest! {
+
+        #[test]
+        fn eval(bytes in collection::vec(any::<u8>(), 33)) {
```
greenhat, line 39: `EcPoint` has `Arbitrary` implemented so the test could be simplified to something like:
```
fn eval(ecp in any::<EcPoint>()) {
    let bytes = ecp.sigma_serialize_bytes();
    let expr: Expr = DecodePoint {input: Expr::Const(bytes.clone().into()).into()}.into();
    let res = eval_out_wo_ctx::<dlog_group::EcPoint>(&expr);
    prop_assert_eq!(res, ecp);
}

File: ergotree-interpreter/src/eval/decode_point.rs [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+use crate::eval::EvalError::Misc;
+use crate::eval::Evaluable;
+use ergotree_ir::mir::constant::TryExtractInto;
+use ergotree_ir::sigma_protocol::dlog_group;
+
+impl Evaluable for DecodePoint {
+    fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let point_bytes = self.input.eval(env, ctx)?.try_extract_into::<Vec<u8>>()?;
+        let point = dlog_group::from_bytes(point_bytes)
+            .map_err(|_| Misc(String::from("Invalid EC point")))?;
```
greenhat, line 17:
```suggestion
            .map_err(|_| Misc(String::from("DecodePoint: Failed to parse EC point from bytes")))?;
```

ergotree-ir/src/mir/decode_point.rs
```diff
@@ -0,0 +1,84 @@
...
+    use crate::mir::expr::arbitrary::ArbExprParams;
+    use proptest::prelude::*;
+
+    impl Arbitrary for DecodePoint {
+        type Strategy = BoxedStrategy<Self>;
+        type Parameters = usize;
+
+        fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+            any_with::<Expr>(ArbExprParams {
+                tpe: SType::SColl(SType::SBoolean.into()),
```
greenhat, line 57:
```suggestion
                tpe: SType::SColl(SType::SByte.into()),
```
?

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -85,6 +85,26 @@ pub fn exponentiate(base: &EcPoint, exponent: &Scalar) -> EcPoint {
     }
 }
 
+/// Attempt to decode
+pub fn from_bytes(bytes: impl AsRef<[u8]>) -> Result<EcPoint, SerializationError> {
```
greenhat, line 89: Do you have a reason to have parsing code outside of `SigmaSerializable` implementation besides test for `eval`? 
`EcPoint::parse_from_bytes(bytes)` will give the same result as this `from_bytes`.

greenhat (changes requested):
Thank you for your contribution. It's looking good. Please take a look at my comments.

greenhat:
> This is first PR for parsing ergo script. In the end we'll need to be able to parse all of ergoscript for ergvein wallet and DecodePoint is as good starting point as any

Thank you! Do you have any shortlist of what IR nodes you need the most? I'll bump them to high priority. 

> I wasn't able to find where DecodePoint is evaluated in scala code so this is only my best guess on how it should be evaluated.

Yes, with current AOT costing the evaluation is quite "buried" in the internal layers. Check out the upcoming JIT costing branch - https://github.com/ScorexFoundation/sigmastate-interpreter/pull/699 . The evaluation is in the IR node class.

> P.S. I noticed that there're quite a few of AST nodes that are just wrappers over `Box<Expr>`. At least `And`, `Or`, `DecodePoint` and probably others. Wouldn't it be better to implement them using single struct parametrized by phantom type parameter to distinguish different nodes? Something along the lines of:
> 
> ```rust
> struct UnaryNode<T> {
>     input: Box<Expr>
> }
> ```

That's an interesting idea. It allows serialization to be covered with a single `SigmaSerializable` implementation. But evaluation, `tpe()` and `op_code()` would still have to be implemented separately for each IR node(`T`). Plus, we have to consider how it affects pattern matching on `Expr` for these nodes. I'd make an experimental PR and tried it on a few IR nodes to see if it worth it.

Shimuuar:
Thanks for quick review!

> Do you have any shortlist of what IR nodes you need the most? I'll bump them to high priority.

No. At the moment I just try to decode some random block and try to add whatever fails to parse

ergotree-ir/src/mir/decode_point.rs
Shimuuar: Copy-paste error.

greenhat:
> > P.S. I noticed that there're quite a few of AST nodes that are just wrappers over `Box<Expr>`. At least `And`, `Or`, `DecodePoint` and probably others. Wouldn't it be better to implement them using single struct parametrized by phantom type parameter to distinguish different nodes? Something along the lines of:
> > ```rust
> > struct UnaryNode<T> {
> >     input: Box<Expr>
> > }
> > ```
> 
> That's an interesting idea. It allows serialization to be covered with a single `SigmaSerializable` implementation. But evaluation, `tpe()` and `op_code()` would still have to be implemented separately for each IR node(`T`). Plus, we have to consider how it affects pattern matching on `Expr` for these nodes. I'd make an experimental PR and tried it on a few IR nodes to see if it worth it.

I created https://github.com/ergoplatform/sigma-rust/issues/228 to try it out.

Shimuuar:
I updated PR and addressed your comments. I also had to extend arbitrary implementation for expressions. Please look at it too

greenhat (approved):
Looking good! Thanks for `Arbitrary` implementation for numeric types!

PR #229, opened 2021-04-08
More types

Shimuuar:
Implementation is copied from  `sigmastate/src/main/scala/sigmastate/types.scala`. Thus being product of copy-paste it's possible that I mixed up something.

P.S. What is good place to put `property` function? So far I just duplicated it

### DIFF
--- a/ergotree-ir/src/types/sglobal.rs
+++ b/ergotree-ir/src/types/sglobal.rs
@@ -37,7 +37,7 @@ lazy_static! {
         method_id: GROUP_GENERATOR_METHOD_ID,
         name: "groupGenerator",
         tpe: SFunc {
-            t_dom: vec![],
+            t_dom: vec![SType::SGlobal],
             t_range: SType::SGroupElement.into(),
             tpe_params: vec![],
         },
@@ -50,8 +50,9 @@ lazy_static! {
         name: "xor",
         tpe: SFunc {
             t_dom: vec![
+                SType::SGlobal,
                 SType::SColl(SType::SByte.into()),
-                SType::SColl(SType::SByte.into())
+                SType::SColl(SType::SByte.into()),
             ],
             t_range: SType::SColl(SType::SByte.into()).into(),
             tpe_params: vec![],
--- a/ergotree-ir/src/types/sglobal.rs
+++ b/ergotree-ir/src/types/sglobal.rs
@@ -1,0 +1,60 @@
+use crate::serialization::types::TypeCode;
+
+use super::sfunc::SFunc;
+use super::smethod::MethodId;
+use super::smethod::SMethodDesc;
+use super::stype::SType;
+use super::stype_companion::STypeCompanion;
+use super::stype_companion::STypeCompanionHead;
+use lazy_static::lazy_static;
+
+/// SGlobal type id
+pub const TYPE_ID: TypeCode = TypeCode::SGLOBAL;
+
+/// groupGenerator property
+pub const GROUP_GENERATOR_METHOD_ID: MethodId = MethodId(1);
+/// "xor" predefined function
+pub const XOR_METHOD_ID: MethodId = MethodId(2);
+
+static S_GLOBAL_TYPE_COMPANION_HEAD: STypeCompanionHead = STypeCompanionHead {
+    type_id: TYPE_ID,
+    type_name: "Global",
+};
+
+lazy_static! {
+    /// Box object type companion
+    pub static ref S_GLOBAL_TYPE_COMPANION: STypeCompanion = STypeCompanion::new(
+        &S_GLOBAL_TYPE_COMPANION_HEAD,
+        vec![
+            &GROUP_GENERATOR_METHOD_DESC,
+            &XOR_METHOD_DESC,
+        ]
+    );
+}
+
+lazy_static! {
+    static ref GROUP_GENERATOR_METHOD_DESC: SMethodDesc = SMethodDesc {
+        method_id: GROUP_GENERATOR_METHOD_ID,
+        name: "groupGenerator",
+        tpe: SFunc {
+            t_dom: vec![],
+            t_range: SType::SGroupElement.into(),
+            tpe_params: vec![],
+        },
+    };
+}
+
+lazy_static! {
+    static ref XOR_METHOD_DESC: SMethodDesc = SMethodDesc {
+        method_id: XOR_METHOD_ID,
+        name: "xor",
+        tpe: SFunc {
+            t_dom: vec![
+                SType::SColl(SType::SByte.into()),
+                SType::SColl(SType::SByte.into())
+            ],
+            t_range: SType::SColl(SType::SByte.into()).into(),
+            tpe_params: vec![],
+        },
+    };
+}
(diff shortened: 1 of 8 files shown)


ergotree-ir/src/types/sglobal.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use super::sfunc::SFunc;
+use super::smethod::MethodId;
+use super::smethod::SMethodDesc;
+use super::stype::SType;
+use super::stype_companion::STypeCompanion;
+use super::stype_companion::STypeCompanionHead;
+use lazy_static::lazy_static;
+
+/// SBox type id
```
greenhat, line 11:
```suggestion
/// SGlobal type id
```

ergotree-ir/src/types/sglobal.rs
```diff
@@ -0,0 +1,60 @@
...
+pub const TYPE_ID: TypeCode = TypeCode::SGLOBAL;
+
+/// Box.groupGenerator property
```
greenhat, line 14:
```suggestion
/// groupGenerator property
```

ergotree-ir/src/types/sglobal.rs
```diff
@@ -0,0 +1,60 @@
...
+pub const GROUP_GENERATOR_METHOD_ID: MethodId = MethodId(1);
+/// Box.xor property
```
greenhat, line 16:
```suggestion
/// "xor" predefined function
```

ergotree-ir/src/types/sglobal.rs
```diff
@@ -0,0 +1,60 @@
...
+lazy_static! {
+    static ref XOR_METHOD_DESC: SMethodDesc = SMethodDesc {
+        method_id: XOR_METHOD_ID,
+        name: "xor",
+        tpe: SFunc {
+            t_dom: vec![
+                SType::SColl(SType::SByte.into()),
+                SType::SColl(SType::SByte.into())
+            ],
+            t_range: SType::SGroupElement.into(),
```
greenhat, line 56:
```suggestion
            t_range: SType::SColl(SType::SByte.into()).into(),
```

ergotree-ir/src/types/sheader.rs
```diff
@@ -0,0 +1,122 @@
...
+lazy_static! {
+    static ref POW_DISTANCE_METHOD: SMethodDesc =
+        property("powDistance", SType::SBigInt, MethodId(14));
+}
+
+lazy_static! {
+    static ref VOTES_METHOD: SMethodDesc = property("votes", SColl(SByte.into()), MethodId(15));
+}
+
+fn property(name: &'static str, tpe: SType, id: MethodId) -> SMethodDesc {
```
greenhat, line 112: It looks like `SMethodDesc` constructor to me. How about moving it into `SMethodDesc::property(...)`?

ergotree-ir/src/types/stype.rs
```diff
@@ -56,6 +56,12 @@ pub enum SType {
     SFunc(SFunc),
     /// Context object ("CONTEXT" in ErgoScript)
     SContext,
+    /// Header of a block
+    SHeader,
+    /// Header o block without solved mining puzzle
```
greenhat, line 61:
```suggestion
    /// Header of a block without solved mining puzzle
```

ergotree-ir/src/types/stype.rs
```diff
@@ -56,6 +56,12 @@ pub enum SType {
...
+    SPreHeader,
+    /// Data type introduced to unify handling of global and non-global (i.e. methods) operations.
+    SGlobal,
```
greenhat, line 64: Please, add new `SType` variants to the `Arbitrary` impl. It'll cover type serialization code.

greenhat (changes requested):
Thank you! Looking good. Please, check my comments.

Shimuuar:
I addressed review comments. 

I also took a second look at `property` function. Now I'm not sure whether it's correct. Should `t_dom` be empty vector or should it be singleton vector? Also current implementation of `PropertyCall` (https://github.com/ergoplatform/sigma-rust/blob/4c269ab6338600955fc1c482404fe74a913ad06c/ergotree-ir/src/mir/property_call.rs#L20-L25) requires that `t_dom` is singleton vector. Is that implementation incomplete?

greenhat:
> I addressed review comments.
> 
> I also took a second look at `property` function. Now I'm not sure whether it's correct. Should `t_dom` be empty vector or should it be singleton vector? Also current implementation of `PropertyCall` (
> 
> https://github.com/ergoplatform/sigma-rust/blob/4c269ab6338600955fc1c482404fe74a913ad06c/ergotree-ir/src/mir/property_call.rs#L20-L25
> 
> ) requires that `t_dom` is singleton vector. Is that implementation incomplete?

Oh, I missed that. `t_dom` should be a singleton vector with an object type of the property (`SGlobal` in case of global/predefined functions). One more parameter to the `property` constructor. Or even better, how about adding `tpe: SType` field to `STypeCompanionHead` and adding `property_tpe` method so that we have a call like `S_BOX_COMPANION_HEAD.property_tpe(...)`?

Shimuuar:
> t_dom should be a singleton vector

That couldn't be right. `xor` should take 2 byte arrays as input as well. Indexing of collections already takes two parameters: 
https://github.com/ergoplatform/sigma-rust/blob/4c269ab6338600955fc1c482404fe74a913ad06c/ergotree-ir/src/types/scoll.rs#L37-L45

Looks like correct invariant is `t_dom` should have at least 1 parameter and first parameter is type of object.

Shimuuar:
P.S. Should `SFunc` of collection indexing mention type variables?

Shimuuar:
> Or even better, how about adding tpe: SType field to STypeCompanionHead

What should it be equal to for collections? Two defined methods use different type variables

greenhat:
> > t_dom should be a singleton vector
> 
> That couldn't be right. `xor` should take 2 byte arrays as input as well. Indexing of collections already takes two parameters:
> https://github.com/ergoplatform/sigma-rust/blob/4c269ab6338600955fc1c482404fe74a913ad06c/ergotree-ir/src/types/scoll.rs#L37-L45
> 
> Looks like correct invariant is `t_dom` should have at least 1 parameter and first parameter is type of object.

Yes, you're right. I should've clarified that singleton vector is only for object properties. For method, it's object's type followed be argument types. `SGlobal` is an exception, only function argument types are in `t_dom`.

greenhat:
> P.S. Should `SFunc` of collection indexing mention type variables?

Yes.

greenhat:
> > Or even better, how about adding tpe: SType field to STypeCompanionHead
> 
> What should it be equal to for collections? Two defined methods use different type variables

`SColl(STypeVar.iv())` should fit ("iv" means "Input Value"). We have this convention in Scala version. And all `Coll` methods use the same type variable (`STypeVar.iv()`) for collection element type in their sigs.

Shimuuar:
I went for variant with extra parameter in property function. Mostly to avoid dealing with question which type should contain `STypeCompanionHead` for collections and why it's different from first parameter of method (different type variables).

greenhat (approved):
Thanks! Sorry for the delay, I did not get from your last comment that this PR was ready for review.

PR #231, opened 2021-04-13
Explicitly handle errors in SMethod::from_ids & and PropertyCall deserialization

Shimuuar:
I stumbled on panics when mass parsing ergo blockchain. I think parsing code should never panic no matter what is thrown at it


greenhat (approved):
Thank you! Good catch! I share the same rule for the parsing code (no panics).

PR #232, opened 2021-04-13
Add MinerPubKey and Global as global variables 

Shimuuar:
This is good enough for parsing existing transactions but doesn't good enough to be merged. 

1. Should global/miner pubkey go to global vars or should they get separate constructor in `Expr`?

2. Scala code is light on documentation. Here is my guess: `MinerPubKey` is public key of miner which created block in which transaction is evaluated. `Global` should just evaluate to sigleton value of type `SGlobal` but there's no corresponding constructor in `Value`

3. Evaluation is absent at the moment.


greenhat:
> 1. Should global/miner pubkey go to global vars or should they get separate constructor in `Expr`?

`MinerPubKey` should be a variant in `GlobalVars`. `Global` seems like deserve it's own constructor in `Expr`. 

> 2. Scala code is light on documentation. Here is my guess: `MinerPubKey` is public key of miner which created block in which transaction is evaluated. `Global` should just evaluate to sigleton value of type `SGlobal` but there's no corresponding constructor in `Value`

Yes, you're right. 
`MinerPubKey` should evaluate to `PreHeader::minerPk` (`Context::pre_header` should be added). 
`Global` should get evaluation similar to `Context`, e.g. new `Value::Global`, module similar to `ergotre_ir::types::scontext` for it's methods/properties, see - 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/sigmastate/types.scala#L2448-L2494

Shimuuar:
Thanks!

> MinerPubKey should evaluate to PreHeader::minerPk (Context::pre_header should be added).

Looks like there's no existing PreHeader data type. There's only dummy in `ergo_lib` which depends on `ergotree-interpreter`

greenhat:
> Thanks!
> 
> > MinerPubKey should evaluate to PreHeader::minerPk (Context::pre_header should be added).
> 
> Looks like there's no existing PreHeader data type. There's only dummy in `ergo_lib` which depends on `ergotree-interpreter`

Yes, please move `PreHeader` to `ergotree-ir`. Here is the Scala version of it for reference - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/sigmastate/eval/CostingDataContext.scala#L436-L444

Shimuuar:
I've implemented everything.

ergotree-ir/src/mir/header.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    impl Arbitrary for PreHeader {
+        type Parameters = ();
+        type Strategy = BoxedStrategy<PreHeader>;
+
+        fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+            (
+                vec(any::<u8>(), 32),
+                // Timestamps between 2000-2050
+                946_674_000_000..2_500_400_300_000u64,
```
greenhat, line 53: Nice!

greenhat (approved):
Thank you! Looking good!

PR #233, opened 2021-04-14
Implement parser and evaluator for DH tuple

Shimuuar:
Another notable change: I made fields of `ergotree-ir::sigma_protocol::sigma_boolean::ProveDhTuple` public


greenhat (approved):
Thanks! Looking good!

PR #234, opened 2021-04-16
Fix flatmap method & GET_VAR opcode

Shimuuar:
GET_VAR should be 115 not 114 [See](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/25251c1313b0131835f92099f02cef8a5d932b5e/sigmastate/src/main/scala/sigmastate/serialization/OpCodes.scala#L181)

Same with method ID of flatmap [See](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/25251c1313b0131835f92099f02cef8a5d932b5e/sigmastate/src/main/scala/sigmastate/types.scala#L1129-L1142). Also it looks like type is wrong. It should be `([i], i->[o]) -> [o]` instead of `([i], i->o) -> [o]`

### DIFF
--- a/ergotree-ir/src/mir/value.rs
+++ b/ergotree-ir/src/mir/value.rs
@@ -63,7 +63,7 @@ impl CollKind {
             SType::SByte => items
                 .into_iter()
                 .map(|v| v.try_extract_into::<i8>())
-                .collect::<Result<Vec<i8>, TryExtractFromError>>()
+                .collect::<Result<Vec<_>, _>>()
                 .map(|bytes| CollKind::NativeColl(NativeColl::CollByte(bytes))),
             _ => Ok(CollKind::WrappedColl { elem_tpe, items }),
         }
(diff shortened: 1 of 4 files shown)


Shimuuar:
I updated evaluator and tests. With `impl StoreWrapped for Value {}` I'm not sure what I'm doing and whether this is right way of dealing with extracting `Vec<Value>`. Concat seems suboptimal too but I wasn't able to find better way. Maybe it does exist, I'm not fluent with rust iterators.

P.S. I stumbled on problem with types when mass parsing transaction on blockchain. With updated type is parses succesfully

ergotree-ir/src/mir/value.rs
```diff
@@ -69,6 +69,28 @@ impl CollKind {
         }
     }
 
+    /// Build a collection from items where each is a collection as well, storing them as Rust types values when neccessary
+    pub fn from_vec_vec(
```
greenhat, line 73: I wonder if `from_vec_vec`(any level of "nestedness" really) could be implemented in `from_vec` with a recursive call. `elem_tpe` will be `SColl(...)`, etc.

greenhat (changes requested):
Thanks! Nice catch! Looking good, please see my comment.

ergotree-ir/src/mir/value.rs
Shimuuar: I don't see any obvious way. Flattening step `Vec<Vec<A>>  -> Vec<A>` must come from somewhere. I'll think about it. I don't like proposed implementation anyway

greenhat: I appreciate your efforts! Let me know if you want me to merge this PR as is.

Shimuuar: Looks like it's not possible to do better with iterators. They don't support chaining of several failible computations. Something like `.map(f:A->Result<B,Err>).map(g:B->Result<C,Err>)`. It's not possible to push error into iterator since `next` returns only `Option` and can't signal error easily.

Only thing I managed to improve is to use partial type hints to guide type inference.

greenhat (approved):
Thanks for your efforts!

PR #235, opened 2021-04-16
Add method descriptions for SContext

Shimuuar:
Some of scontext's methods use [different functions](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/25251c1313b0131835f92099f02cef8a5d932b5e/sigmastate/src/main/scala/sigmastate/types.scala#L1716-L1728) to define properties. I don't understand difference so I defined all properties in the same way


greenhat (approved):
Thanks! Looks good!

PR #236, opened 2021-04-19
Implement GetVar

Shimuuar:
This is partial implementation of GetVar. Serialization is implemented but there's  only skeleton for evaluator.

 1. What types are allowed in GetVar? For now tests just hardcode `SInt`.
 2. Evaluation should apparently extract some variables from context. But what are those variables? Are they not present in context yet?

PR also contains fix to SigmaToBoolean type


greenhat:
> 1. What types are allowed in GetVar? For now tests just hardcode `SInt`.

Any type data type, really (any `Constant` to be precise).  

> 2. Evaluation should apparently extract some variables from context. But what are those variables? Are they not present in context yet?
Yes, there is `ContextExtension` that holds the variables. Could you add it to `Context`, please. Here is an eval for `getVar` - 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/sigmastate/eval/CostingDataContext.scala#L749-L783
Be aware, the variables at the end are stored in an array with gaps for missing ids. See - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala#L133-L140
I'm quite sure it was done for performance reasons only. We can skip it and keep using hash map for storage.

Shimuuar:
I added simplest possible set of variables. Should it be `HashMap<u8,Constant>` or `HashMap<u8,Value>`? I picked former since it looks there's no way to compute type of `Value`

Set of variables in context is always initialized to empty map. Is this right choice for `ergo_lib::wallet::signing::make_context`? What is good choice for `Arbitrary` instance for `Context`?

ergotree-interpreter/src/eval/context.rs
```diff
@@ -23,6 +25,8 @@ pub struct Context {
     pub inputs: Vec<IrBoxId>,
     /// Pre header of current block
     pub pre_header: PreHeader,
+    ///
+    pub var_map: HashMap<u8, Constant>,
```
greenhat, line 29: Could you make it  `extension: ContextExtension,` instead, please? We're going to need to manipulate multiple `ContextExtension` in the future, and to keep it consistent with Scala version (see `ErgoLikeContext`, our `Context` sister type there).

ergotree-interpreter/src/eval/get_var.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::eval::Evaluable;
+use ergotree_ir::mir::constant::TryExtractFromError;
+
+impl Evaluable for GetVar {
+    fn eval(&self, _env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        match ctx.ctx.var_map.get(&self.var_id) {
+            None => Ok(Value::Opt(None.into())),
+            Some(v) if v.tpe == self.var_tpe => Ok((Some(v.v.clone())).into()),
+            Some(v) => Err(TryExtractFromError(format!(
+                "expected {:?}, found {:?}",
```
greenhat, line 17:
```suggestion
                "GetVar: expected {:?}, found {:?}",
```

ergotree-interpreter/src/eval/get_var.rs
```diff
@@ -0,0 +1,23 @@
...
+                self.var_tpe, v.tpe
+            ))
+            .into()),
+        }
+    }
+}
```
greenhat, line 23: Although the eval code is fairly simple it'd be nice to have tests covering all three cases, just to catch a regression.

greenhat (changes requested):
Thanks! Looking good!

> I added simplest possible set of variables. Should it be `HashMap<u8,Constant>` or `HashMap<u8,Value>`? I picked former since it looks there's no way to compute type of `Value`.

Good choice! But `ContextExtension` is even better (see my comment).
 
> Set of variables in context is always initialized to empty map. Is this right choice for `ergo_lib::wallet::signing::make_context`? 

In the future, it should be passed to `make_context` as a parameter. Let's use empty for now (`ContextExtension::empty()`).

> What is good choice for `Arbitrary` instance for `Context`?

`ContextExtension` has `Arbitrary` implementation. Let's pass that.

Shimuuar:
I've addressed eveything

ergotree-interpreter/src/eval/context.rs
```diff
@@ -23,6 +24,8 @@ pub struct Context {
     pub inputs: Vec<IrBoxId>,
     /// Pre header of current block
     pub pre_header: PreHeader,
+    ///
```
greenhat, line 27:
```suggestion
    /// prover-defined key-value pairs, that may be used inside a script
```

greenhat (approved):

greenhat:
> I've addressed eveything

Thank you!

PR #237, opened 2021-04-19
Implement Atleast AST node

Shimuuar:
It only covers serialization of AST. I think it's better to avoid touching sigma expressions while AND/OR connectives  are under development.


greenhat (approved):
Thanks! Looking good.

PR #239, opened 2021-04-20
Implementation of Deserialize{Context,Register}

Shimuuar:
It's still partial implementation. There's no evaluation and types could be wrong. 

I'm still confused how it should be evaluated and wasn't able to find corresponding code in scala version

### DIFF
--- a/ergotree-ir/src/mir/deserialize_context.rs
+++ b/ergotree-ir/src/mir/deserialize_context.rs
@@ -1,0 +1,82 @@
+//! Extracts context variable as byte array
+
+use crate::serialization::op_code::OpCode;
+use crate::serialization::sigma_byte_reader::SigmaByteRead;
+use crate::serialization::sigma_byte_writer::SigmaByteWrite;
+use crate::serialization::SerializationError;
+use crate::serialization::SigmaSerializable;
+use crate::types::stype::SType;
+
+/// Extracts context variable as byte array
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct DeserializeContext {
+    /// Type of variable
+    pub ctx_tpe: SType,
+    /// Context ID
+    pub ctx_id: u8,
+}
+impl DeserializeContext {
+    pub(crate) const OP_CODE: OpCode = OpCode::DESERIALIZE_CONTEXT;
+
+    /// FIXME: Is this true??
+    pub fn tpe(&self) -> SType {
+        self.ctx_tpe.clone()
+    }
+
+    pub(crate) fn op_code(&self) -> OpCode {
+        Self::OP_CODE
+    }
+}
+
+impl SigmaSerializable for DeserializeContext {
+    fn sigma_serialize<W: SigmaByteWrite>(&self, w: &mut W) -> Result<(), std::io::Error> {
+        self.ctx_tpe.sigma_serialize(w)?;
+        w.put_u8(self.ctx_id)
+    }
+
+    fn sigma_parse<R: SigmaByteRead>(r: &mut R) -> Result<Self, SerializationError> {
+        let ctx_tpe = SType::sigma_parse(r)?;
+        let ctx_id = r.get_u8()?;
+        Ok(Self { ctx_tpe, ctx_id })
+    }
+}
+
+#[cfg(feature = "arbitrary")]
+/// Arbitrary impl
+mod arbitrary {
+    use super::*;
+
+    use proptest::prelude::*;
+
+    impl Arbitrary for DeserializeContext {
+        type Strategy = BoxedStrategy<Self>;
+        type Parameters = ();
+
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            (0_u8..9, any::<SType>())
+                .prop_map(|(ctx_id, ctx_tpe)| Self { ctx_tpe, ctx_id })
+                .boxed()
+        }
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "arbitrary")]
+mod tests {
+    use crate::mir::expr::Expr;
+    use crate::serialization::sigma_serialize_roundtrip;
+
+    use super::*;
+
+    use proptest::prelude::*;
+
+    proptest! {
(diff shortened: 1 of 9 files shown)


ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
+//! Extracts context variable as byte array
+
+use crate::serialization::op_code::OpCode;
+use crate::serialization::sigma_byte_reader::SigmaByteRead;
+use crate::serialization::sigma_byte_writer::SigmaByteWrite;
+use crate::serialization::SerializationError;
+use crate::serialization::SigmaSerializable;
+use crate::types::stype::SType;
+
+/// Extracts context variable as byte array
```
greenhat, line 10:
```suggestion
/// Extracts context variable as Coll[Byte], deserializes it to script and then executes this script in the current context.
/// The original `Coll[Byte]` of the script is available as `getVar[Coll[Byte]](id)`
/// On evaluation returns the result of the script execution in the current context
```

ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct DeserializeContext {
+    /// Type of variable
```
greenhat, line 13:
```suggestion
    /// Result type of the deserialized script
```

ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub ctx_tpe: SType,
```
greenhat, line 14:
```suggestion
    pub tpe: SType,
```

ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Context ID
```
greenhat, line 15:
```suggestion
    /// identifier of the context variable
```

ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub ctx_id: u8,
```
greenhat, line 16:
```suggestion
    pub id: u8,
```

ergotree-ir/src/mir/deserialize_context.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+impl DeserializeContext {
+    pub(crate) const OP_CODE: OpCode = OpCode::DESERIALIZE_CONTEXT;
+
+    /// FIXME: Is this true??
```
greenhat, line 21: Yes, it's the result type of the deserialized script execution.

ergotree-ir/src/mir/deserialize_register.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::serialization::sigma_byte_writer::SigmaByteWrite;
+use crate::serialization::SerializationError;
+use crate::serialization::SigmaSerializable;
+use crate::types::stype::SType;
+
+/// Extract register of SELF box as byte array, deserialize it into Value and inline into executing script.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct DeserializeRegister {
+    /// Register number
+    pub reg_n: u8,
```
greenhat, line 15:
```suggestion
    pub reg_id: u8,
```

ergotree-ir/src/mir/deserialize_register.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Type of value in register
+    pub reg_tpe: SType,
```
greenhat, line 17:
```suggestion
    pub tpe: SType,
```

ergotree-ir/src/mir/deserialize_register.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// FIXME:
+    pub reg_val: Option<Box<Expr>>,
```
greenhat, line 19:
```suggestion
    pub default: Option<Box<Expr>>,
```

ergotree-ir/src/mir/deserialize_register.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+impl DeserializeRegister {
+    pub(crate) const OP_CODE: OpCode = OpCode::DESERIALIZE_REGISTER;
+
+    /// FIXME: Is this true??
```
greenhat, line 24: Yes. The type of value stored in the register.

greenhat:
Thanks!

 > I'm still confused how it should be evaluated and wasn't able to find corresponding code in scala version

Yes, these IR nodes are more complicated.

They both are handled before the script execution. See - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala#L85-L102
and
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeInterpreter.scala#L21-L40
 
I think we can skip their evaluation for now. I have not thought about it yet, feel free to take a stab at it if you want though.

Regarding the names. We're trying to have consistent naming between Scala and Rust versions. So, if code translates "directly" its better to use Scala names.

Shimuuar:
I updated field names to mirror scala better. 

> I think we can skip their evaluation for now. I have not thought about it yet, feel free to take a stab at it if you want though.

Let skip evaluation then. After all main main goal at the moment to be able to parse whole ergo blockchain

greenhat (approved):
Great! There is one commit suggestion (comment change) left unaccepted.

Shimuuar:
Yep. Missed that one. Updated

greenhat:
@Shimuuar Resolve the merge conflict(after merged GetVar), please.

Shimuuar:
Done.

Issue #228, opened 2021-04-07
Try `UnaryNode<T>` for "single value" IR nodes


greenhat:
from #227

> > P.S. I noticed that there're quite a few of AST nodes that are just wrappers over `Box<Expr>`. At least `And`, `Or`, `DecodePoint` and probably others. Wouldn't it be better to implement them using single struct parametrized by phantom type parameter to distinguish different nodes? Something along the lines of:
> > ```rust
> > struct UnaryNode<T> {
> >     input: Box<Expr>
> > }
> > ```
> 
> That's an interesting idea. It allows serialization to be covered with a single `SigmaSerializable` implementation. But evaluation, `tpe()` and `op_code()` would still have to be implemented separately for each IR node(`T`). Plus, we have to consider how it affects pattern matching on `Expr` for these nodes. I'd make an experimental PR and tried it on a few IR nodes to see if it worth it.

PR #241, opened 2021-04-21
Implement few AST nodes as UnaryNode parameterized by phantom type

Shimuuar:
I need to add few more nodes which only has single `Expr` argument. I don't want to write too much boilerplate so I decide to try out approach from #228.

It turns out rust doesn't like phantom type parameters and forces one to insert `PhantomData<T>` into data type definition. Still it worked out reasonably well. It gives simple API for implementation of different nodes, ensures that everything is treated uniformly. @greenhat what's your opinion?


greenhat:
Thank you for trying this out! I appreciate your effort.

Serialization came out great! But the price we pay by complicating things worries me. I mean comparing it with an existing approach

```rust
#[derive(PartialEq, Eq, Debug, Clone)]
pub struct CalcBlake2b256 {
    pub input: Box<Expr>,
}

impl CalcBlake2b256 {
    const OP_CODE: OpCode = OpCode::CALC_BLAKE2B256;

    pub fn new(input: Expr) -> Result<Self, InvalidArgumentError> {
        input.check_post_eval_tpe(SType::SColl(Box::new(SType::SByte)))?;
        Ok(CalcBlake2b256 {
            input: Box::new(input),
        })
    }

    pub fn tpe(&self) -> SType {
        SType::SColl(Box::new(SType::SByte))
    }
}
```

vs. Unary<T>:

```rust
#[derive(PartialEq, Eq, Debug, Clone)]
pub struct TagCalcBlake2b256;

pub type CalcBlake2b256 = Unary<TagCalcBlake2b256>;

impl TagOpCode for TagCalcBlake2b256 {
    const OP_CODE: OpCode = OpCode::CALC_BLAKE2B256;

    fn res_tpe(_: &Expr) -> SType {
        SType::SColl(Box::new(SType::SByte))
    }

    fn check_arg(expr: &Expr) -> Result<(), InvalidArgumentError> {
        expr.check_post_eval_tpe(SType::SColl(Box::new(SType::SByte)))?;
        Ok(())
    }
}
```

Given that we can replace `op_code()` in the existing approach with some `HasOpCode` trait that would add a line `impl HasOpCode for CalcBlake2b256 {}` for each node.
```rust
trait HasOpCode{
    const OP_CODE: OpCode;

    fn op_code(&self) -> OpCode {
        Self::OP_CODE
    }
}
```

The Unary<T> approach does not look simpler to me and would require the understanding of the "Unary<T> + TagOpCode + TagNode + Node newtype" hierarchy.
I've tried to find more cases where we'd benefit from unification (like the serialization) and could not find any so far. Do you see any?
Also, I'm wondering how ergonomics of Unary<T> will look like in the compiler(frontend) with a lot of pattern matching and rewriting trees (variable binding, type inference, etc.). 
I also see `TagOpCode` holds more than opcode machinery but also `check_arg`. So it either should be split into multiple traits or renamed to something more generic like `TagUnaryNode`, which does not simplify overall code comprehension.


Alternative idea - traits that allow blanket serialization implementation for one-value-input nodes. 

```rust
pub trait UnaryOp {
    fn input(&self) -> &Expr;
}

pub trait CheckedCtor {
    fn new(expr: Expr) -> Result<Self, InvalidArgumentError>;
}

impl<T: UnaryOp, CheckedCtor> SigmaSerializable for T {
    fn sigma_serialize<W: SigmaByteWrite>(&self, w: &mut W) -> Result<(), std::io::Error> {
        self.input().sigma_serialize(w)
    }

    fn sigma_parse<R: SigmaByteRead>(r: &mut R) -> Result<Self, SerializationError> {
        let input = Expr::sigma_parse(r)?;
        let r = T::new(input)?;
        Ok(r)
    }
}
```

So for every node that implements `UnaryOp` and `CheckedCtor` we would have serialization implemented automatically.
What do you think?

Shimuuar:
> Also, I'm wondering how ergonomics of Unary will look like in the compiler(frontend) with a lot of pattern matching and rewriting trees (variable binding, type inference, etc.).

This I think kills idea outright.`phantom` field will get in the way and will be generally unpleasant to work with. Also there could be possible problems with type inference. So `UnaryNode` should be scrapped,

> Alternative idea - traits for existing nodes that allow blanket serialization implementation for one-value-input nodes.

This will work. However problem with blanket implementation is you only can have one. I'm not sure it's _the_ one. It's probably better to live with duplication at least until AST is complete

greenhat:
> > Also, I'm wondering how ergonomics of Unary will look like in the compiler(frontend) with a lot of pattern matching and rewriting trees (variable binding, type inference, etc.).
> 
> This I think kills idea outright.`phantom` field will get in the way and will be generally unpleasant to work with. Also there could be possible problems with type inference. So `UnaryNode` should be scrapped,

I agree. I foresee some difficulties there as well.
 
> > Alternative idea - traits for existing nodes that allow blanket serialization implementation for one-value-input nodes.
> 
> This will work. However problem with blanket implementation is you only can have one. I'm not sure it's _the_ one. It's probably better to live with duplication at least until AST is complete

The beauty of this binary serialization is that it's a part of the consensus protocol and will not be changed in the future.

PR #244, opened 2021-04-23
implementation of nodes for conversion between byte array and numbers

Shimuuar:
Namely:

 - ByteArrayToLong
 - ByteArrayToBigInt
 - LongToByteArray

There's only stub implementation for evaluator since I wasn't able to find exact semantics for it


greenhat:
> There's only stub implementation for evaluator since I wasn't able to find exact semantics for it

Here is Scala version:
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6c51c13f7a494a191a7ea5645e56b04fb46a418d/sigma-impl/src/main/scala/special/sigma/SigmaDslOverArrays.scala#L56-L65

For Long <-> byte array conversion it calls:
``` java
  /**
   * Returns a big-endian representation of {@code value} in an 8-element byte array; equivalent to
   * {@code ByteBuffer.allocate(8).putLong(value).array()}. For example, the input value
   * {@code 0x1213141516171819L} would yield the byte array {@code {0x12, 0x13, 0x14, 0x15, 0x16,
   * 0x17, 0x18, 0x19}}.
   *
   * <p>If you need to convert and concatenate several values (possibly even of different types),
   * use a shared {@link java.nio.ByteBuffer} instance, or use
   * {@link com.google.common.io.ByteStreams#newDataOutput()} to get a growable buffer.
   */
  public static byte[] toByteArray(long value) {
    // Note that this code needs to stay compatible with GWT, which has known
    // bugs when narrowing byte casts of long values occur.
    byte[] result = new byte[8];
    for (int i = 7; i >= 0; i--) {
      result[i] = (byte) (value & 0xffL);
      value >>= 8;
    }
    return result;
  }

  /**
   * Returns the {@code long} value whose big-endian representation is stored in the first 8 bytes
   * of {@code bytes}; equivalent to {@code ByteBuffer.wrap(bytes).getLong()}. For example, the
   * input byte array {@code {0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19}} would yield the
   * {@code long} value {@code 0x1213141516171819L}.
   *
   * <p>Arguably, it's preferable to use {@link java.nio.ByteBuffer}; that library exposes much more
   * flexibility at little cost in readability.
   *
   * @throws IllegalArgumentException if {@code bytes} has fewer than 8 elements
   */
  public static long fromByteArray(byte[] bytes) {
    checkArgument(bytes.length >= BYTES, "array too small: %s < %s", bytes.length, BYTES);
    return fromBytes(
        bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7]);
  }

  /**
   * Returns the {@code long} value whose byte representation is the given 8 bytes, in big-endian
   * order; equivalent to {@code Longs.fromByteArray(new byte[] {b1, b2, b3, b4, b5, b6, b7, b8})}.
   *
   * @since 7.0
   */
  public static long fromBytes(
      byte b1, byte b2, byte b3, byte b4, byte b5, byte b6, byte b7, byte b8) {
    return (b1 & 0xFFL) << 56
        | (b2 & 0xFFL) << 48
        | (b3 & 0xFFL) << 40
        | (b4 & 0xFFL) << 32
        | (b5 & 0xFFL) << 24
        | (b6 & 0xFFL) << 16
        | (b7 & 0xFFL) << 8
        | (b8 & 0xFFL);
  }
```

ergotree-ir/src/serialization/expr.rs
```diff
@@ -214,6 +220,9 @@ impl SigmaSerializable for Expr {
                 SigmaPropBytes::OP_CODE => Ok(SigmaPropBytes::sigma_parse(r)?.into()),
                 Tuple::OP_CODE => Ok(Tuple::sigma_parse(r)?.into()),
                 DecodePoint::OP_CODE => Ok(DecodePoint::sigma_parse(r)?.into()),
+                OpCode::BYTE_ARRAY_TO_LONG => Ok(ByteArrayToLong::sigma_parse(r)?.into()),
```
greenhat, line 223: I've used `OpCode::XXX` format in the beginning, but later I found `Node::OP_CODE` format more readable. Plus, the latter format "glues" IR node with its opcode.

greenhat (changes requested):
Thanks! Looking good! Please see the comments.

Shimuuar:
I added evaluators for long<->byte arrays. Spec as I understood it is:

- Long is serialized as 2-complement 8 bytes BE.
- Decoded as 8 bytes BE and all bytes after first 8 are ignored. Less than 8 bytes is decode error.

How bigint should be parsed? I assume it's 2-complement big-endian too. What is acceptable range? -2^255 — 2^255-1?

greenhat:
> I added evaluators for long<->byte arrays. Spec as I understood it is:
> 
> * Long is serialized as 2-complement 8 bytes BE.
> * Decoded as 8 bytes BE and all bytes after first 8 are ignored. Less than 8 bytes is decode error.

This sound right.

> How bigint should be parsed? I assume it's 2-complement big-endian too. What is acceptable range? -2^255 — 2^255-1?

Yes, in scala it's: 
```scala
    val bi = new BigInteger(bytes.toArray).to256BitValueExact
```

which calls:

```java
    /**
     * Translates a byte sub-array containing the two's-complement binary
     * representation of a BigInteger into a BigInteger.  The sub-array is
     * specified via an offset into the array and a length.  The sub-array is
     * assumed to be in <i>big-endian</i> byte-order: the most significant
     * byte is the element at index {@code off}.  The {@code val} array is
     * assumed to be unchanged for the duration of the constructor call.
     *
     * An {@code IndexOutOfBoundsException} is thrown if the length of the array
     * {@code val} is non-zero and either {@code off} is negative, {@code len}
     * is negative, or {@code off+len} is greater than the length of
     * {@code val}.
     *
     * @param  val byte array containing a sub-array which is the big-endian
     *         two's-complement binary representation of a BigInteger.
     * @param  off the start offset of the binary representation.
     * @param  len the number of bytes to use.
     * @throws NumberFormatException {@code val} is zero bytes long.
     * @throws IndexOutOfBoundsException if the provided array offset and
     *         length would cause an index into the byte array to be
     *         negative or greater than or equal to the array length.
     * @since 9
     */
    public BigInteger(byte[] val, int off, int len) {
        if (val.length == 0) {
            throw new NumberFormatException("Zero length BigInteger");
        }
        Objects.checkFromIndexSize(off, len, val.length);

        if (val[off] < 0) {
            mag = makePositive(val, off, len);
            signum = -1;
        } else {
            mag = stripLeadingZeroBytes(val, off, len);
            signum = (mag.length == 0 ? 0 : 1);
        }
        if (mag.length >= MAX_MAG_LENGTH) {
            checkRange();
        }
    }

    /**
     * Translates a byte array containing the two's-complement binary
     * representation of a BigInteger into a BigInteger.  The input array is
     * assumed to be in <i>big-endian</i> byte-order: the most significant
     * byte is in the zeroth element.  The {@code val} array is assumed to be
     * unchanged for the duration of the constructor call.
     *
     * @param  val big-endian two's-complement binary representation of a
     *         BigInteger.
     * @throws NumberFormatException {@code val} is zero bytes long.
     */
    public BigInteger(byte[] val) {
        this(val, 0, val.length);
    }
```

and then, checks if it fits:

```scala
    /** Checks this {@code BigInteger} can be cust to 256 bit two's-compliment representation,
      * checking for lost information. If the value of this {@code BigInteger}
      * is out of the range of the 256 bits, then an {@code ArithmeticException} is thrown.
      *
      * @return this {@code BigInteger} if the check is successful
      * @throws ArithmeticException if the value of {@code this} will
      * not exactly fit in a 256 bit range.
      * @see BigInteger#longValueExact
      */
    @inline final def to256BitValueExact: BigInteger = {
      // Comparing with 255 is correct because bitLength() method excludes the sign bit.
      // For example, these are the boundary values:
      // (new BigInteger("80" + "00" * 31, 16)).bitLength() = 256
      // (new BigInteger("7F" + "ff" * 31, 16)).bitLength() = 255
      // (new BigInteger("-7F" + "ff" * 31, 16)).bitLength() = 255
      // (new BigInteger("-80" + "00" * 31, 16)).bitLength() = 255
      // (new BigInteger("-80" + "00" * 30 + "01", 16)).bitLength() = 256
      if (x.bitLength() <= 255) x
      else
        throw new ArithmeticException("BigInteger out of 256 bit range");
    }
  }
```

Shimuuar:
I added BigInt evaluator and tests. Bit count doesn't work same for rust's BigInt and -2^256 will get bit count of 256, so I had to resort to use constants.

ergotree-interpreter/src/eval/byte_array_to_bigint.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let input = self.input.eval(env, ctx)?.try_extract_into::<Vec<u8>>()?;
+        if input.is_empty() {
+            return Err(UnexpectedValue(
+                "ByteArrayToBigInt: byte array is empty".into(),
+            ));
+        }
+        let n = num_bigint::BigInt::from_signed_bytes_be(&input);
+        println!("{}", n);
+        println!("{:x}", n);
```
greenhat, line 25: I assume these println!() calls were for debugging purposes and are no longer needed.

greenhat (changes requested):
Thank you! Looking good! I appreciate your efforts to make sure we have evaluation semantics in sync with the Scala version. Please see my comment about println calls.

ergotree-interpreter/src/eval/byte_array_to_bigint.rs
Shimuuar: Yes. They were to test that tests test what they supposed to test

Shimuuar:
I also added test that makes sure that empty array is decode error

greenhat:
> I also added test that makes sure that empty array is decode error

Great! Thanks!

greenhat (approved):

PR #251, opened 2021-04-29
Use LSB bitvector instead of MSB

### LINKED ISSUE #249: Transacrion in block 400188 fails to roundtrip
Namely [this one](https://explorer.ergoplatform.com/en/transactions/9da0ea07e4f859b6a71817d9ccb67f43ddfa1f8a59f7cf2af18a544bcd799706). Offending ergo tree is: `1000d1ed8503` which deserialized as
```
BoolToSigmaProp(
    BoolToSigmaProp {
        input: BinOp(
            BinOp {
                kind: Relation(
                    And,
                ),
                left: Const(
                    Constant {
                        tpe: SBoolean,
                        v: Boolean(false,),
                    },
                ),
                right: Const(
                    Constant {
                        tpe: SBoolean,
                        v: Boolean(false,),
                    },
                ),
            },
        ),
    },
)
```

I suspect problem is in `ReadSigmaVlqExt::get_bits`. I it decodes bit vector in MST while it should use LSB instead.

Shimuuar:
sigma_ser::vlq_encode::...::{put,get}_bits should use LSB bit vectors.
Fix required switch from bit-vec to bitvec which supports LSB vectors.

Fixes #249, regression test added

### DIFF
--- a/ergotree-ir/src/mir/bin_op.rs
+++ b/ergotree-ir/src/mir/bin_op.rs
@@ -201,7 +201,11 @@ mod arbitrary {
 mod tests {
 
     use super::*;
+    use crate::mir::constant::Constant;
+    use crate::mir::expr::Expr;
+    use crate::mir::value::Value::Boolean;
     use crate::serialization::sigma_serialize_roundtrip;
+    use crate::serialization::SigmaSerializable;
     use proptest::prelude::*;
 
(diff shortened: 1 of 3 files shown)


greenhat (approved):
Thank you! It looks great! I appreciate the regression test.

PR #256, opened 2021-05-12
introduce HasOpCode trait;

### LINKED ISSUE #243: DRY idea for IR node's `fn op_code(&self)`
Extracted from #241

> Given that we can replace op_code() in the existing approach with some HasOpCode trait that would add a line impl HasOpCode for CalcBlake2b256 {} for each node.

```rust
trait HasOpCode{
    const OP_CODE: OpCode;

    fn op_code(&self) -> OpCode {
        Self::OP_CODE
    }
}
```

greenhat:
Close #243

### DIFF
--- a/ergotree-ir/src/has_opcode.rs
+++ b/ergotree-ir/src/has_opcode.rs
@@ -1,9 +1,15 @@
 use crate::serialization::op_code::OpCode;
 
-pub(crate) trait HasOpCode {
+pub(crate) trait HasStaticOpCode {
     const OP_CODE: OpCode;
+}
 
+pub(crate) trait HasOpCode {
+    fn op_code(&self) -> OpCode;
+}
+
+impl<T: HasStaticOpCode> HasOpCode for T {
     fn op_code(&self) -> OpCode {
-        Self::OP_CODE
+        T::OP_CODE
     }
 }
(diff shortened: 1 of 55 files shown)


greenhat:
@Shimuuar Hey! Check out this `HasOpCode` trait. Seems like a good fit.

Shimuuar:
I like that approach because we do have some sort of interface: this value has corresponding opcode. Main problem is:

```rust
   const OP_CODE: OpCode;
```

For some some nodes opcode could be inferred purely from type like `CalcBlake2b256` so it's possible to define it. While for `BinOp` opcode depends on actual binary operation. One way to solve it is to reduce trait to:

```rust
pub(crate) trait HasOpCode {
    fn op_code(&self) -> OpCode;
}
```

But it becomes impossible to pattern match on OP_CODE. In principle it should be possible to define two traits, but I'm not sure whether rust allows to define default trait implementation in terms of another trait which may or may not be defined.

greenhat:
Good point. Although we exclude `BinOp` and similar IR nodes from implementing this trait I wonder if something like this will work for both "static" and "dynamic" op code nodes:

```rust
pub trait HasStaticOpCode {
    const OP_CODE: OpCode;
}

pub trait HasOpCode {
    fn op_code(&self) -> OpCode;
}

impl<T: HasStaticOpCode> HasOpCode for T {
    fn op_code(&self) -> OpCode {
        T::OP_CODE
    }
}
```
So that all nodes implementing `HasStaticOpCode` get `HasOpCode` implementation automatically. `BinOp` and other will implement `HasOpCode` manually.

Shimuuar:
Yes. I think that should work. That's what meant by "define two traits"

PR #257, opened 2021-05-12
introduce UnaryOp and UnaryOpTryBuild traits + SigmaSerializable impl;

### LINKED ISSUE #242: DRY idea for SigmaSerializable impl for one-paramater IR nodes
Extracted from #241  

> Alternative idea - traits for existing nodes that allow blanket serialization implementation for one-value-input nodes.

```rust
pub trait UnaryOp {
    fn input(&self) -> &Expr;
}

pub trait CheckedCtor {
    fn new(expr: Expr) -> Result<Self, InvalidArgumentError>;
}

impl<T: UnaryOp, CheckedCtor> SigmaSerializable for T {
    fn sigma_serialize<W: SigmaByteWrite>(&self, w: &mut W) -> Result<(), std::io::Error> {
        self.input().sigma_serialize(w)
    }

    fn sigma_parse<R: SigmaByteRead>(r: &mut R) -> Result<Self, SerializationError> {
        let input = Expr::sigma_parse(r)?;
        let r = T::new(input)?;
        Ok(r)
    }
}
```
> So for every node that implements UnaryOp and CheckedCtor we would have serialization implemented automatically.

greenhat:
Close #242

- [x] use traits in the rest of unary op IR nodes;

### DIFF
--- a/ergotree-interpreter/src/eval/coll_filter.rs
+++ b/ergotree-interpreter/src/eval/coll_filter.rs
@@ -86,6 +86,7 @@ mod tests {
     use ergotree_ir::mir::func_value::FuncArg;
     use ergotree_ir::mir::func_value::FuncValue;
     use ergotree_ir::mir::property_call::PropertyCall;
+    use ergotree_ir::mir::unary_op::UnaryOpTryBuild;
     use ergotree_ir::mir::val_use::ValUse;
     use ergotree_ir::types::scontext;
     use ergotree_ir::types::stype::SType;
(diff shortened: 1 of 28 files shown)


greenhat:
@Shimuuar I took a stab at serialization DRY for unary ops and use it in a couple of nodes. What do you think?

Shimuuar:
It looks quite reasonable to me. Only minor nitpick. Is following necessary in trait implementations?

```
    where
        Self: Sized,
```

rust should be able to infer that Self is sized since it's known there

greenhat:
Good catch, thanks! I eliminated it by making `pub trait UnaryOpTryBuild: Sized`.

PR #260, opened 2021-05-16
sigma_parse_bytes does not need mutable vector as parameter

Shimuuar:
And could work quite happily with just &[u8]. This allowed to remove few
conversions but mostly amounted to just slapping & everywhere.

### DIFF
--- a/ergo-lib/src/chain/json.rs
+++ b/ergo-lib/src/chain/json.rs
@@ -39,7 +39,8 @@ pub mod ergo_tree {
         String::deserialize(deserializer)
             .and_then(|str| base16::decode(&str).map_err(|err| Error::custom(err.to_string())))
             .and_then(|bytes| {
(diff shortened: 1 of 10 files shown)


ergo-lib/src/chain/json.rs
```diff
@@ -129,17 +130,17 @@ pub mod ergo_box {
     impl TryFrom<Base16DecodedBytes> for ConstantWrapper {
         type Error = ConstantParsingError;
 
-        fn try_from(bytes: Base16DecodedBytes) -> Result<Self, Self::Error> {
-            let c = Constant::sigma_parse_bytes(bytes.into())?;
+        fn try_from(Base16DecodedBytes(bytes): Base16DecodedBytes) -> Result<Self, Self::Error> {
```
greenhat, line 133: I did not know that you can de-structure in parameters. Neat!

greenhat (approved):
Thanks for catching this!

PR #271, opened 2021-05-21
Remove dbg! calls

### LINKED ISSUE #258: Clean up `dbg!()` calls.
It should not be used for prolonged periods of time in non-test code.

Shimuuar:
Attempt to parse blockchain data result in torrent in debug print. 

This PR simply removes all `dbg!` calls. debug_tree must be defined only for test build since otherwise it trips
dead_code lint:
    
> error: associated function is never used: `debug_tree`

Fixes #258

### DIFF
--- a/ergo-lib/src/chain/json.rs
+++ b/ergo-lib/src/chain/json.rs
@@ -283,7 +283,6 @@ mod tests {
         #[test]
         fn tx_roundtrip(t in any::<Transaction>()) {
             let j = serde_json::to_string(&t)?;
-            // dbg!(j);
             eprintln!("{}", j);
             let t_parsed: Transaction = serde_json::from_str(&j)?;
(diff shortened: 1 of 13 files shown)


greenhat:
@Shimuuar Thanks for taking care of this! 
I see you removed `dbg!` calls from tests as well. I thought `dbg!` should not bother anyone in tests.
I'd rather leave `dbg!` in tests to have their output since it does not "pollute" the log unless the test is failing. Plus, although we don't compile tests with release profile I anxious that the compiler might eliminate "unused" values.

Shimuuar:
I restored all `dbg!` where rust complained about unused variables. I think that removing panic would be invalid optimization but it's better to be on safe side.

greenhat (approved):
Looking great! Thanks!

PR #272, opened 2021-05-21
Add SigmaSerializable for Box<T>

Shimuuar:
Reasoning is simple: `Box` is just wrapper type which changes its memory layout, so in serialization code it could be just ignored. This allows to write more generic implementation for `Option<T>` instead if `Option<Box<T>>`. As small bonus this allows write parsing code in style:

```rust
   let field = <_>::sigma_parse(r)?;
```
And let type inference to do its work


Shimuuar:
Ah. This rather old patch conflicts with `impl<T: UnaryOp + UnaryOpTryBuild> SigmaSerializable for T`

PR #277, opened 2021-05-27
Remove Peekable constraint from ReadSigmaVlqExt

Shimuuar:
ReadSigmaVlqExt does not use Peekable anywhere
    
    So here we downgrade constraint from Peekable to mere Read and move Peekable to
    SigmaByteRead. It doesn't strictly need SigmaByteRead either but it's used in
    SigmaSerializable which requires it.
    
    Main motivation is parsing of network messages which use VLQ encoding quite a
    bit but does not require ability to peek. In fact only parsing of ergo script
    requires it.


greenhat (approved):
Looking great! Thanks!

PR #280, opened 2021-05-31
Base16-based implementation of Debug for Digest32

Shimuuar:
This PR adds base-16 Debug instance for Digest2. It's way more useful this way since it much more compact and allows to compare it directly with block explorer data, search, etc.

It also contain few other unrelated changes

1. Use from `#from` annotation to derive From instance
2. Show error text for IO error
3. Use plain `[u8;3]` array for `Votes`. No reason to add pointer for 3-bytes array.


greenhat (approved):
Nice! Thanks!

PR #284, opened 2021-06-02
Remove Peekable trait and PeekableReader

Shimuuar:
Turns out need for peeking is quite localized and only `Expr`, `Constant` and `SType` need special treatment. 

@greenhat What's your opinion? This is feasibility test so I skipped on writing comments


greenhat:
Hey! It's looking great!

greenhat (approved):
Looking great! Thanks!

PR #287, opened 2021-06-04
Start making definition of block header complete

Shimuuar:
This draft PR adds all fields for block header except for the proof of PoW. Few outstanding questions:

1. `ADDigest` looks almost same as `Digest32`. Only difference is length. One way to unify them is to use const generics and make digest length parameter and then make `type Digest32 = Digest<32>`

2. Autolykos proof contain EC point which require serde traint implementations. Machinery for deriving base16-based serde instances is located in ergo lib and not accessible for `EcPoint`. Maybe it should be moved to `sigma-ser` to be accessible everywhere.

### DIFF
--- a/ergo-lib/src/chain.rs
+++ b/ergo-lib/src/chain.rs
@@ -9,6 +9,8 @@ mod digest32;
 pub use base16_bytes::*;
 pub use digest32::*;
 
+///
+pub mod addigest;
 pub mod block_header;
 pub mod contract;
 pub mod ergo_box;
--- a/ergo-lib/src/chain/addigest.rs
+++ b/ergo-lib/src/chain/addigest.rs
@@ -1,0 +1,105 @@
+use crate::chain::{Base16DecodedBytes, Base16EncodedBytes};
+use ergotree_ir::util::AsVecI8;
+#[cfg(feature = "json")]
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use thiserror::Error;
+
+///
+#[cfg_attr(feature = "json", derive(Serialize, Deserialize))]
+#[cfg_attr(
+    feature = "json",
+    serde(into = "Base16EncodedBytes", try_from = "Base16DecodedBytes")
+)]
+#[derive(PartialEq, Eq, Hash, Debug, Clone)]
+pub struct ADDigest(pub Box<[u8; Self::SIZE]>);
+
+impl ADDigest {
+    /// Digest size 33 bytes
+    pub const SIZE: usize = 33;
+}
+
+impl From<[u8; ADDigest::SIZE]> for ADDigest {
+    fn from(bytes: [u8; ADDigest::SIZE]) -> Self {
+        ADDigest(Box::new(bytes))
+    }
+}
+
+impl From<ADDigest> for Base16EncodedBytes {
+    fn from(v: ADDigest) -> Self {
+        Base16EncodedBytes::new(v.0.as_ref())
+    }
+}
+
+impl From<ADDigest> for Vec<i8> {
+    fn from(v: ADDigest) -> Self {
+        v.0.to_vec().as_vec_i8()
+    }
+}
+
+impl From<ADDigest> for Vec<u8> {
+    fn from(v: ADDigest) -> Self {
+        v.0.to_vec()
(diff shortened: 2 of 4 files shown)


greenhat:
> This draft PR adds all fields for block header except for the proof of PoW. Few outstanding questions:
> 
> 1. `ADDigest` looks almost same as `Digest32`. Only difference is length. One way to unify them is to use const generics and make digest length parameter and then make `type Digest32 = Digest<32>`

I very much agree! Const generics seems like an excellent fit here.

> 2. Autolykos proof contain EC point which require serde traint implementations. Machinery for deriving base16-based serde instances is located in ergo lib and not accessible for `EcPoint`. Maybe it should be moved to `sigma-ser` to be accessible everywhere.

From what I understand you intend to use one header type for IR and ergo-lib. 
I don't think we should have one header type for all. 
JSON encoding should not be in the IR layer so I'd rather keep it contained in `ergo-lib` and above. 
Besides that header type used in the interpreter 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/89afefa935d52dd83741f6530f97e4c935992f76/sigma-api/src/main/scala/special/sigma/SigmaDsl.scala#L650-L658
 
differs from 
Ergo block header - https://github.com/ergoplatform/ergo/blob/937ab21555f1962dd140e7168f26bb61a9ddd896/src/main/scala/org/ergoplatform/modifiers/history/Header.scala#L75-L87

Shimuuar:
> JSON encoding should not be in the IR layer so I'd rather keep it contained in ergo-lib and above.

I'm afraid it's not possible because of traits. Trait implementation must be defined together with data type definition. Which again brings us to `EcPoint`. If it implements serde traits it must be done in `ergotree-ir`. If it doesn't it results in weird situation where header has instances defined and some of its fields doesn't.

With trait based APIs I think it's better to go all in define trait implementations for all types where it makes sense. 

> Besides that header type used in the interpreter differs from Ergo block header

They _are_ different but they are almost same:

1. Interpreter version has block ID but it computed from header proper and could be turned into method that use memoization to avoid recomputing it.
2. It has V1 pow solution hardcoded. I think V1 solution could be trivially derived from V2.
3. Biggest one is it contains full avltree instead of just hash. Maybe it could be worked around by adding type parameter and some type level magic. I'm not sure it will work nicely or even at all

Whether juice worth the squeeze is another question

greenhat:
> > JSON encoding should not be in the IR layer so I'd rather keep it contained in ergo-lib and above.
> 
> I'm afraid it's not possible because of traits. Trait implementation must be defined together with data type definition. Which again brings us to `EcPoint`. If it implements serde traits it must be done in `ergotree-ir`. If it doesn't it results in weird situation where header has instances defined and some of its fields doesn't.
> 
> With trait based APIs I think it's better to go all in define trait implementations for all types where it makes sense.

There is a feature "remote derive" https://serde.rs/remote-derive.html which I used for `ContextExtension` 
https://github.com/ergoplatform/sigma-rust/blob/9ac04dc0204061d422cdf6e656b1535ee4b72a4e/ergo-lib/src/chain/json/context_extension.rs#L7-L16
and `ProofBytes`. 
 
> > Besides that header type used in the interpreter differs from Ergo block header
> 
> They _are_ different but they are almost same:
> 
> 1. Interpreter version has block ID but it computed from header proper and could be turned into method that use memoization to avoid recomputing it.
> 2. It has V1 pow solution hardcoded. I think V1 solution could be trivially derived from V2.
> 3. Biggest one is it contains full avltree instead of just hash. Maybe it could be worked around by adding type parameter and some type level magic. I'm not sure it will work nicely or even at all
> 
> Whether juice worth the squeeze is another question

I see the interpreter's header as some kind of a "projection" of the chain's block header. They serve slightly different purposes. Checkout how ErgoBox is presented in IR - https://github.com/ergoplatform/sigma-rust/blob/9ac04dc0204061d422cdf6e656b1535ee4b72a4e/ergotree-ir/src/ir_ergo_box.rs#L43-L62
`IrErgoBox` trait is that "projection" that we need from ErgoBox in IR and in the interpreter. 
I'm not sure that trait is needed for the header though, maybe, a struct would suffice. I chose to use trait for the box to avoid copying the data on every box access operation which is quite popular. Header access OTOH is a far less popular operation.

Issue #288, opened 2021-06-04
Ergotree BigInt values are not always checked for overflow


jerry-best:
According to the language spec, Ergotree BigInts behave as a 256-bit signed two's complement integer, and should produce an error on any operation which over/under-flows. They're implemented using an arbitrary-precision big integer library in rust, so arithmetic operations which produce a value which can't be represented in 256 bits don't actually trigger an error when they should. In the scala implementation, these operations look like they do check and error out as appropriate.

The behavior of the arithmetic operators in ergotree-interpreter/src/eval/bin_op.rs should be changed to issue an error on overflow. For example, the following test currently fails, but should pass:

`
   #[test]
   fn test_mult_overflow() {
        let big : BigInt = 1.to_bigint().unwrap() << 250;
        assert!(eval_num_op::<BigInt>(ArithOp::Multiply, big.clone().into(), big.clone().into()).is_err());
    }
`

Issue #289, opened 2021-06-06
Ergotree numeric negation may not handle overflow properly


jerry-best:
For all the signed integer types, negation can overflow (when the value is the minimum value). As currently implemented in ergotree-ir/src/types/stype.rs , this will result in either a panic (debug build), or a wrong answer (release build). This should probably be changed to return Error instead.

PR #294, opened 2021-06-07
Fix issues #288 and #289: Make bigint operations check for 256-bit overflow and Change numeric negation on all types to give an error on overflow.

jerry-best:
Make negation return an error on overflow for all integral types (including BigInt). This happens when negating the minimum value.

Separately, (but included in this PR because it depends on the previous change), check for overflow on the arithmetic operations on BigInt (+-*/) and error in that case. I considered a more involved change of adding a 256-bit integer type (implemented using bigint, and implementing the required numeric traits) which would check on all ops, but this would require replacing use of BigInt everywhere, and seemed a bit invasive.

### DIFF
--- a/ergotree-interpreter/src/eval/bin_op.rs
+++ b/ergotree-interpreter/src/eval/bin_op.rs
@@ -34,8 +34,8 @@ fn arithmetic_err<T: std::fmt::Display>(
     ))
 }
 
-fn check_bigint_overflow(res: Result<Value, EvalError>) -> Result<Value, EvalError> {
-    let bigint = res?.try_extract_into::<BigInt>()?;
+fn check_bigint_overflow(val: Value) -> Result<Value, EvalError> {
+    let bigint = val.try_extract_into::<BigInt>()?;
     if fits_in_256_bits(&bigint) {
         Ok(bigint.into())
     } else {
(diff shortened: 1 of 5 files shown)


ergotree-interpreter/src/eval/bin_op.rs
```diff
@@ -16,65 +16,77 @@ use num_traits::CheckedSub;
...
+    rv_raw: T,
+    err_str: &str,
+) -> EvalError {
+    EvalError::ArithmeticException(format!(
+        "({0}) {1} ({2}) resulted in {3}",
+        lv_raw, op, rv_raw, err_str
+    ))
+}
+
+fn check_bigint_overflow(res: Result<Value, EvalError>) -> Result<Value, EvalError> {
```
greenhat, line 37: Consider reducing `res` type to `Value` and use `?` before passing an argument at the call site. 
After that renaming `res` to `v` might be a better fit.

greenhat (changes requested):
Looking great! Please check a minor comment.

greenhat:
> ... I considered a more involved change of adding a 256-bit integer type (implemented using bigint, and implementing the required numeric traits) which would check on all ops, but this would require replacing use of BigInt everywhere, and seemed a bit invasive.

Actually, it's a very good idea! I'm going to make an issue for this. Feel free to jump on it.

ergotree-interpreter/src/eval/bin_op.rs
jerry-best: Sounds good, just made that change and I'll push it shortly.

greenhat (approved):
Great! Thanks for the quick response!

Issue #299, opened 2021-06-08
[100 SigmaUSD] Implement `CalcSha256` op for byte arrays.


greenhat:
Create a new struct in a separate module(file) in `mir` module and add it to `Expr` enum. 
Implement `UnaryOp` and `UnaryOpTryBuild`.

### Serialization
- Implement `Arbitrary` trait and add serialization roundtrip PBT test for this node;

### Evaluation
- Implement `Evaluable` trait in `ergotree-interpreter` crate;
- Add evaluation tests;

Check out the existing IR node implementation (`CalcBlake2b256`) for details.

Check out the detailed implementation instructions for `Append` op in  #301

Reference Scala version:
IR node and eval - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/89afefa935d52dd83741f6530f97e4c935992f76/sigmastate/src/main/scala/sigmastate/trees.scala#L719-L737

qy3u:
Awesome guide!I'll try this implemention.

greenhat:
> Awesome guide!I'll try this implemention.

Merged. Thank you for your contribution! Please contact me (my username is `greenhat`) on our Discord https://discord.gg/kj7s7nb to receive your bounty.

qy3u:
My discord username is qyu. :)

PR #303, opened 2021-06-10
Implement CalcSha256 op for byte arrays

qy3u:
This is my Implemention for https://github.com/ergoplatform/sigma-rust/issues/299
I hope it's a correct implemention.If there are any errors, please remind me!

### DIFF
--- a/ergotree-ir/src/serialization/expr.rs
+++ b/ergotree-ir/src/serialization/expr.rs
@@ -13,6 +13,7 @@ use crate::mir::bool_to_sigma::BoolToSigmaProp;
 use crate::mir::byte_array_to_bigint::ByteArrayToBigInt;
 use crate::mir::byte_array_to_long::ByteArrayToLong;
 use crate::mir::calc_blake2b256::CalcBlake2b256;
+use crate::mir::calc_sha256::CalcSha256;
 use crate::mir::coll_by_index::ByIndex;
 use crate::mir::coll_exists::Exists;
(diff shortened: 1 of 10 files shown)


ergotree-ir/src/serialization/expr.rs
```diff
@@ -133,6 +134,7 @@ impl Expr {
                 ExtractAmount::OP_CODE => Ok(Expr::ExtractAmount(ExtractAmount::sigma_parse(r)?)),
                 OpCode::SELECT_FIELD => Ok(Expr::SelectField(SelectField::sigma_parse(r)?)),
                 OpCode::CALC_BLAKE2B256 => Ok(CalcBlake2b256::sigma_parse(r)?.into()),
+                OpCode::CALC_SHA256 => Ok(CalcSha256::sigma_parse(r)?.into()),
```
greenhat, line 137:
```suggestion
                CalcSha256::OP_CODE => Ok(CalcSha256::sigma_parse(r)?.into()),
```
To reuse node op code “binding”. Plus, an eye catches it quickly to re-affirm op code to node association here.

greenhat (changes requested):
Welcome and thanks for your contribution! It’s looking great! See one minor suggestion.

ergotree-ir/src/serialization/expr.rs
qy3u: Thank you for your advice! Updated.

greenhat (approved):

PR #304, opened 2021-06-10
Fix for UnsignedTransaction's from-JSON conversion

rooooooooob:
After tx ids were removed from UnsignedTransaction, the automatic serde
deserialization no longer worked as there was no way for it to know how
to construct the tx_id field. This just does an approach similar to how
Transaction does its JSON conversions.

We could have alternatively just removed the tx_id field entirely and
computed it whenever we needed it, but I wasn't sure how this would
impact other users efficiency-wise so I went with this approach instead.

### DIFF
--- a/bindings/ergo-lib-wasm/src/transaction.rs
+++ b/bindings/ergo-lib-wasm/src/transaction.rs
@@ -74,6 +74,13 @@ impl Transaction {
         JsValue::from_serde(&self.0.clone()).map_err(|e| JsValue::from_str(&format!("{}", e)))
     }
 
+    /// JSON representation
+    pub fn from_json(json: &str) -> Result<Transaction, JsValue> {
+        serde_json::from_str(json)
+            .map(Self)
+            .map_err(|e| JsValue::from_str(&format!("{}", e)))
+    }
+
     /// Inputs for transaction
     pub fn inputs(&self) -> Inputs {
         self.0.inputs.clone().into()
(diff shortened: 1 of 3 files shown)


ergo-lib/src/chain/json/transaction.rs
```diff
@@ -18,6 +18,21 @@ pub struct TransactionJson {
...
+    #[cfg_attr(feature = "json", serde(rename = "dataInputs"))]
+    pub data_inputs: Vec<DataInput>,
+    /// box candidates to be created by this transaction
+    #[cfg_attr(feature = "json", serde(rename = "outputs"))]
+    pub outputs: Vec<ErgoBoxCandidate>,
+}
+
 #[cfg(test)]
 mod tests {
     use crate::chain::transaction::unsigned::UnsignedTransaction;
```
rooooooooob, line 38: Curiously, the test `unsigned_tx_roundtrip()` passes beforehand, but when we tried to actually deserialize from JSON an unsigned transaction (from the wasm bindings but those use the same `serde_json::from_str()` method) it complains that there's a missing field `tx_id`.

bindings/ergo-lib-wasm/src/transaction.rs
```diff
@@ -74,6 +74,13 @@ impl Transaction {
         JsValue::from_serde(&self.0.clone()).map_err(|e| JsValue::from_str(&format!("{}", e)))
     }
 
+    /// JSON representation
```
rooooooooob, line 77: I just added this for consistency's sake since I was looking through this files while debugging this problem and `UnsignedTransaction` had both but `Transaction` only had `to_json()`

rooooooooob:
I'm not sure why the github build failed on `use std::convert::TryFrom;` being an "unused" import in `unsigned.rs`, but rustc doesn't compile it locally without it as it doesn't know where TryFrom is. Is this a version difference between my local builds and the test jobs? I'm on 1.52.1 (I think this is the most recent)

greenhat (changes requested):
It's looking good! Thank you! I suggested a fix for the failed default-features-disabled build.

greenhat (approved):

PR #306, opened 2021-06-10
Make Digest32 a const-generic 

Shimuuar:
This PR turns `Digest32` into `Digest<N>` that is parameterized by constant. All instances are provided for generic versions. 

`Digest32` and freshly `ADDigest` are defined as:
```rust 
pub type Digest32 = Digest<32>;
pub type ADDigest = Digest<33>;
```
Note that it's not possible to use `sigma_util::DIGEST32_SIZE` in this definition. `ADDigest` is not documented since I couldn't find any sensible documentation on its use.

### DIFF
--- a/ergo-lib/src/chain/digest32.rs
+++ b/ergo-lib/src/chain/digest32.rs
@@ -25,7 +25,7 @@ pub struct Digest<const N: usize>(pub Box<[u8; N]>);
 /// Usually this is as blake2b hash of serialized form
 pub type Digest32 = Digest<32>;
 
-/// 33-byte digest
+/// AVL tree digest: root hash along with tree height (33 bytes)
 pub type ADDigest = Digest<33>;
 
 impl<const N: usize> Digest<N> {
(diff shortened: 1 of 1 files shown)


ergo-lib/src/chain/digest32.rs
```diff
@@ -14,98 +12,104 @@ use std::fmt::Formatter;
...
-#[cfg_attr(test, derive(Arbitrary))]
-pub struct Digest32(pub Box<[u8; Digest32::SIZE]>);
+pub struct Digest<const N: usize>(pub Box<[u8; N]>);
+
+/// 32 byte array used as ID of some value: block, transaction, etc.
+/// Usually this is as blake2b hash of serialized form
+pub type Digest32 = Digest<32>;
 
-impl Digest32 {
+/// 33-byte digest
```
greenhat, line 28:
```suggestion
/// AVL tree digest: root hash along with tree height (33 bytes)
```

greenhat (approved):
It looks awesome! Thank you! I suggested the doc comment for ADDigest.

greenhat (approved):

PR #309, opened 2021-06-11
Implement Slice op for collection

FitzOReilly:


ergotree-ir/src/mir/coll_by_index.rs
```diff
@@ -30,7 +30,7 @@ impl ByIndex {
         let input_elem_type: SType = *match input.post_eval_tpe() {
             SType::SColl(elem_type) => Ok(elem_type),
             _ => Err(InvalidArgumentError(format!(
-                "Expected Map input to be SColl, got {0:?}",
+                "Expected ByIndex input to be SColl, got {0:?}",
```
greenhat, line 33: Thanks for spotting this!

ergotree-ir/src/mir/coll_slice.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ok(Self {
+            input: input.into(),
+            from: from.into(),
+            until: until.into(),
+        })
+    }
+
+    /// Type
+    pub fn tpe(&self) -> SType {
+        match self.input.post_eval_tpe() {
```
greenhat, line 56: The `Slice` result type is the same as `input`, so `self.input.tpe()` should be ok here.

greenhat (changes requested):
It's looking great! See minor comment on `tpe()`. I appreciate extensive eval tests!

ergotree-ir/src/mir/coll_by_index.rs
FitzOReilly: You're welcome!

ergotree-ir/src/mir/coll_slice.rs
FitzOReilly: Fixed

ergotree-ir/src/mir/coll_slice.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ok(Self {
+            input: input.into(),
+            from: from.into(),
+            until: until.into(),
+        })
+    }
+
+    /// Type
+    pub fn tpe(&self) -> SType {
+        match self.input.tpe() {
```
greenhat, line 56: I meant `Slice::tpe()` should not return the element type of the collection, but the "whole" collection type (`SColl(T)`). 
Since the result collection type is the same as `input` we can simple return `self.input.tpe()`. So the method would look like:
```rust
pub fn tpe(&self) -> SType {
    self.input.tpe()
}
```

greenhat (changes requested):

ergotree-ir/src/mir/coll_slice.rs
FitzOReilly: Ok, now I get it. So this method is supposed to return `SColl(T)`, but (wrongly) returned just `T`, if I understand it correctly what you're saying (because a slice is a collection and not a single element). Thanks for clarifying this! Changed.

greenhat: Yes, exactly!

greenhat (approved):

PR #310, opened 2021-06-12
Byte-wise XOR for byte arrays

ApexTheory:
Here is my submission for the byte-array XOR op.  Please provide your feedback; I expect some improvements as I am just getting familiar with the project.

### DIFF
--- a/ergotree-interpreter/src/eval/xor.rs
+++ b/ergotree-interpreter/src/eval/xor.rs
@@ -1,15 +1,14 @@
-use ergotree_ir::mir::xor::Xor;
 use ergotree_ir::mir::value::CollKind;
 use ergotree_ir::mir::value::NativeColl;
 use ergotree_ir::mir::value::Value;
+use ergotree_ir::mir::xor::Xor;
 
 use crate::eval::env::Env;
 use crate::eval::EvalContext;
 use crate::eval::EvalError;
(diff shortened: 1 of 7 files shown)


ergotree-interpreter/src/eval/xor.rs
```diff
@@ -0,0 +1,145 @@
...
+use ergotree_ir::mir::value::NativeColl;
+use ergotree_ir::mir::value::Value;
+use ergotree_ir::mir::xor::Xor;
+
+use crate::eval::env::Env;
+use crate::eval::EvalContext;
+use crate::eval::EvalError;
+use crate::eval::Evaluable;
+
+fn _helper_xor(mut x: Vec<i8>, y: Vec<i8>) -> Vec<i8> {
```
greenhat, line 11: What is the reason for the underscore prefix? I mean, there is  "silence an unused var warning" meaning employed in rustic that first comes to mind.

greenhat (commented):
It's looking great! Thank you! See a minor note. I appreciate extensive eval tests!

ergotree-interpreter/src/eval/xor.rs
ApexTheory: Just an oversight, it has been fixed in the latest commit.

greenhat (approved):

PR #312, opened 2021-06-14
Allow access to ValId & export ConstantPlaceholder

### LINKED ISSUE #307: Visibility of data types.
Few questions arising from attempt to parse transactions from blockchain:

1. `pub struct ValId(u32)` has private field and no accessors. Since there're no contraints on value of underlying `u32` I think field should be just made public.

2. Is there reason to include `pub struct SigmaProp(SigmaBoolean)`? Naively plain `SigmaBoolean` would work as well.

3. `ergotree_ir::mir::constant::ConstantPlaceholder` is only visible to crate. Since it's part of `Expr` I believe it's an oversight.

Shimuuar:
Fixes #307

### DIFF
--- a/ergotree-ir/src/mir/constant.rs
+++ b/ergotree-ir/src/mir/constant.rs
@@ -13,7 +13,7 @@ use std::convert::TryInto;
 
 mod constant_placeholder;
 
-pub(crate) use constant_placeholder::*;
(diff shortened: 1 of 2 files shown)


greenhat (approved):
Than you for taking care of this!

PR #316, opened 2021-06-16
298 - MultiplyGroup op for GroupElement

ApexTheory:
Please see my solution for the MultiplyGroup op and provide your feedback.  
I was hoping to add some eval tests, but I wasn't able to effectively instantiate the test EcPoint values.  If you have some suggestions on that, I would be happy to add some further tests.

### DIFF
--- a/ergotree-interpreter/src/eval.rs
+++ b/ergotree-interpreter/src/eval.rs
@@ -60,6 +60,7 @@ pub(crate) mod if_op;
 pub(crate) mod logical_not;
 pub(crate) mod long_to_byte_array;
 pub(crate) mod method_call;
+pub(crate) mod multiply_group;
 pub(crate) mod negation;
 pub(crate) mod option_get;
 pub(crate) mod option_get_or_else;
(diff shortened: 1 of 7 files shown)


greenhat:
> Please see my solution for the MultiplyGroup op and provide your feedback.
> I was hoping to add some eval tests, but I wasn't able to effectively instantiate the test EcPoint values. If you have some suggestions on that, I would be happy to add some further tests.

Thank you! It looks great! 
For `EcPoint` values check out https://github.com/ergoplatform/sigma-rust/blob/3b1b3a5cee6fdafa5366d76f0e94902eea6fa223/ergotree-ir/src/sigma_protocol/dlog_group.rs#L65-L74
for group generator and identity values along with inverse operation. There is also `DlogProverInput::random().public_image().h` to get random `EcPoint`.

greenhat:
Since `Arbitrary` impl for `EcPoint` includes only group generator and identity value it seems reasonable to add a random `EcPoint` value.
Something like `any::<DlogProverInput>().prop_map(|i| i.public_image().h)`.

greenhat:
Wait! Scratch out everything I wrote above. :) `DlogProverInput` is in `ergotree-interpreter` crate and it cannot be accessed in `ergotree-ir` crate (deps are other way around). We probably need to move `DlogProverInput` to `ergotree-ir` crate, but it should be outside the scope of this PR.

Regarding 
> I was hoping to add some eval tests, but I wasn't able to effectively instantiate the test EcPoint values. If you have some suggestions on that, I would be happy to add some further tests.

I think the existing eval proptest with `EcPoint` is enough.

greenhat (approved):
Looks great! Thank you!

PR #318, opened 2021-06-18
Implement SubstConst node

Shimuuar:
I'm not quite sure what does it do. So there's no meaningful comment. 

1. What should `tpe` return for `SubstConstants`? To infer type one must decode script and to type check it. 
2. Should parsing check that types of arguments are correct?

### DIFF
--- a/ergotree-interpreter/src/eval/expr.rs
+++ b/ergotree-interpreter/src/eval/expr.rs
@@ -11,6 +11,7 @@ impl Evaluable for Expr {
         ctx.cost_accum.add_cost_of(self)?;
         match self {
             Expr::Const(c) => Ok(c.v.clone()),
+            Expr::SubstConstants(_) => todo!(),
             Expr::ByteArrayToLong(op) => op.eval(env, ctx),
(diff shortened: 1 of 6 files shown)


greenhat:
> I'm not quite sure what does it do. So there's no meaningful comment.

It's a way to change constants in a serialized ErgoTree (`script_bytes`). I cannot remember the use cases. Think of it as if you need to compare known script bytes to a script "parametrized" with different constant (PK, token id, etc.).

> 
>     1. What should `tpe` return for `SubstConstants`? To infer type one must decode script and to type check it.

`Coll[Byte]`

> 
>     2. Should parsing check that types of arguments are correct?

Ideally, yes. `script_bytes` should be `Coll[Byte]`, `positions` should be `Coll[Int]` and `new_values` is skipped.

Scala version for reference - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/89afefa935d52dd83741f6530f97e4c935992f76/sigmastate/src/main/scala/sigmastate/trees.scala#L739-L791

Shimuuar:
@greenhat  I've implemented your suggestions

PR #321, opened 2021-06-20
297 - Exponentiate GroupElement

ApexTheory:
Here is my attempt at the Exponentiate op for GroupElement.  

I tried to do the BigInt -> Scalar conversion in a cleaner way using try_into(), but was not able to get it to work.  If you have any suggestions on that, I can make improvements.

### DIFF
--- a/ergotree-interpreter/src/eval/exponentiate.rs
+++ b/ergotree-interpreter/src/eval/exponentiate.rs
@@ -15,7 +15,15 @@ impl Evaluable for Exponentiate {
         let right_v = self.right.eval(env, ctx)?;
 
         let right_bui: Option<BigUint> = match right_v.clone() {
-            Value::BigInt(bi) => BigInt::to_biguint(&bi),
+            Value::BigInt(bi) => match bi.sign() {
+                num_bigint::Sign::Minus => {
+                    return Err(EvalError::UnexpectedValue(format!(
+                        "Expected Exponentiate op exponent to be positive, got: {0:?}",
+                        bi
+                    )))
+                }
+                _ => BigInt::to_biguint(&bi),
+            },
             _ => None,
         };
 
@@ -41,12 +49,12 @@ impl Evaluable for Exponentiate {
 mod tests {
     use super::*;
     use crate::eval::context::Context;
-    use crate::eval::tests::eval_out;
+    use crate::eval::tests::{eval_out, try_eval_out};
+    use crate::sigma_protocol::private_input::DlogProverInput;
 
     use ergotree_ir::mir::expr::Expr;
     use ergotree_ir::sigma_protocol::dlog_group::EcPoint;
-    use num_bigint::RandBigInt;
-    use num_traits::Signed;
+    use num_bigint::{BigInt, Sign};
     use proptest::prelude::*;
     use sigma_test_util::force_any_val;
     use std::rc::Rc;
@@ -54,10 +62,12 @@ mod tests {
     proptest! {
 
         #[test]
-        fn eval_any(left in any::<EcPoint>()) {
+        fn eval_any(left in any::<EcPoint>(), pi in any::<DlogProverInput>()) {
 
-            let mut rng = rand::thread_rng();
-            let right = rng.gen_bigint(64).abs();
+            let r_g_array = pi.w.to_bytes();
+            let r_b_array: &[u8] = r_g_array.as_slice();
+
+            let right: BigInt = BigInt::from_bytes_le(Sign::Plus, r_b_array);
 
             let expected_exp = dlog_group::exponentiate(
                 &left,
@@ -72,6 +82,24 @@ mod tests {
 
             let ctx = Rc::new(force_any_val::<Context>());
             assert_eq!(eval_out::<EcPoint>(&expr, ctx), expected_exp);
+        }
+
+        #[test]
+        fn eval_negative_exponent(left in any::<EcPoint>(), pi in any::<DlogProverInput>()) {
+
+            let r_g_array = pi.w.negate().to_bytes();
+            let r_b_array: &[u8] = r_g_array.as_slice();
+
+            let right: BigInt = BigInt::from_bytes_le(Sign::Minus, r_b_array);
+
+            let expr: Expr = Exponentiate {
+                left: Box::new(Expr::Const(left.into())),
+                right: Box::new(Expr::Const(right.into())),
+            }
+            .into();
+
+            let ctx = Rc::new(force_any_val::<Context>());
+            assert!(try_eval_out::<EcPoint>(&expr, ctx).is_err());
         }
     }
 }
--- a/ergotree-interpreter/src/eval/exponentiate.rs
+++ b/ergotree-interpreter/src/eval/exponentiate.rs
@@ -27,12 +27,12 @@ impl Evaluable for Exponentiate {
             _ => None,
         };
 
-        let right_scalar: Option<Scalar> = match right_bui.clone() {
+        let right_scalar: Option<Scalar> = match right_bui {
             Some(bui) => dlog_group::from_biguint(bui),
             _ => None,
         };
 
-        match (left_v.clone(), right_scalar.clone()) {
+        match (left_v.clone(), right_scalar) {
             (Value::GroupElement(group), Some(exp)) => {
                 Ok(dlog_group::exponentiate(&group, &exp).into())
             }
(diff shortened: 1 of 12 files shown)


ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -105,6 +107,20 @@ pub fn random_scalar_in_group_range() -> Scalar {
...
 
+/// Attempts to create Scalar from BigUint
+/// Returns None if not in the range [0, modulus).
+pub fn from_biguint(b: BigUint) -> Option<Scalar> {
+    let bytes_be = b.to_bytes_be();
+    let bytes = bytes_be.as_slice().clone();
+
+    let mut bytes_32 = [0; 32];
+    for (i, v) in bytes.clone().iter().enumerate() {
+        bytes_32[i] = *v;
```
greenhat, line 118: `bytes` can be larger than 32 bytes since `BigUint` is not bounded.

EDIT: I mean although `BigInt` in ErgoScript is 256-bit we cannot presume that input always fit in 256 bits.

ergotree-interpreter/src/eval/exponentiate.rs
```diff
@@ -0,0 +1,77 @@
...
+    use sigma_test_util::force_any_val;
+    use std::rc::Rc;
+
+    proptest! {
+
+        #[test]
+        fn eval_any(left in any::<EcPoint>()) {
+
+            let mut rng = rand::thread_rng();
+            let right = rng.gen_bigint(64).abs();
```
greenhat, line 60: Seems like a new version of `get_random` broke Wasm build (see CI). I think we can avoid the need of random BigInt here by using random `Scalar` (add `pi in any::<DlogProverInput>()` and use `pi.w`) and converting it to BigInt to use in exp op. And to check the case with out-of-bound BigInt in a separate test.

greenhat (changes requested):
Looking good! Thank you! Please, see the comments.

greenhat:
> Here is my attempt at the Exponentiate op for GroupElement.
Thank you for your contribution!
 
> I tried to do the BigInt -> Scalar conversion in a cleaner way using try_into(), but was not able to get it to work. If you have any suggestions on that, I can make improvements.

Yeah, I feel you. Unfortunately, there is not much we can do about it. No-orphans rule is a double-edge sword.

ergotree-interpreter/src/eval/exponentiate.rs
ApexTheory: I made this change, seems to be working.  Was able to add the negative exponent test case using pi.w.negate().

ergotree-interpreter/src/eval/exponentiate.rs
```diff
@@ -0,0 +1,105 @@
...
+                right: Box::new(Expr::Const(right.into())),
+            }
+            .into();
+
+            let ctx = Rc::new(force_any_val::<Context>());
+            assert_eq!(eval_out::<EcPoint>(&expr, ctx), expected_exp);
+        }
+
+        #[test]
+        fn eval_negative_exponent(left in any::<EcPoint>(), pi in any::<DlogProverInput>()) {
```
greenhat, line 88: I'm not sure we need a PBT for negative BigInt. I mean one value/case should suffice. 
`force_any_val::<EcPoint>()` can be used to get a random value.

Also, it would be nice to cover BigInt > 256 bits. One test with hardcoded value (e.g. `BigInt::parse_bytes(b"2240553423075396383515373673723462837462821468959827293462897346923874293642946928374923875928657983456938759287459236459287459238759346592837", 10).unwrap()` should suffice.

greenhat (changes requested):
Looking good! Thank you! Please, check the comments.

ApexTheory:
Updated as per your latest comments:
1) BigInt parsed from big-endian byte array
2) Negative exponent test moved out of proptest
3) Exponent >256 bit test added

Please see my one comment about the negative exponent mod operation in the Scala code.

ergotree-interpreter/src/eval/exponentiate.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    proptest! {
+
+        #[test]
+        fn eval_any(left in any::<EcPoint>(), pi in any::<DlogProverInput>()) {
+
+            let r_g_array = pi.w.to_bytes();
+            let r_b_array: &[u8] = r_g_array.as_slice();
+
+            let right: BigInt = BigInt::from_bytes_be(Sign::Plus, r_b_array);
```
greenhat, line 70: Seems like it worth to extract into `dlog_group::scalar_to_bigint` to avoid future be/le confusions. 
To unify naming `from_biguint` can be renamed to `bigint_to_scalar` with integrated check for positive values.

greenhat (changes requested):
It's looking great! Thank you! I have one more suggestion. 🙂
I don't see comment regarding Scala code anywhere.

ApexTheory:
Updated with changes for your latest feedback.

Regarding the scala code comment, it is during the check for the positive exponent:

If the exponent is negative, I am just returning an Error. However, in the Scala code they perform this mod operation:
val exp = if (exponent.compareTo(BigInteger.ZERO) < 0) exponent.mod(order) else exponent

Should I be using something like the mod_floor() method available in BigInt for similar functionality?

greenhat:
> Updated with changes for your latest feedback.

Thank you! I appreciate your efforts!
 
> Regarding the scala code comment, it is during the check for the positive exponent:
> 
> If the exponent is negative, I am just returning an Error. However, in the Scala code they perform this mod operation:
> `val exp = if (exponent.compareTo(BigInteger.ZERO) < 0) exponent.mod(order) else exponent`
> 
> Should I be using something like the mod_floor() method available in BigInt for similar functionality?

Hey! What a great find! Actually, the exact eval code in Scala is different(without the check for negative exponent) and I missed this difference. I created https://github.com/ScorexFoundation/sigmastate-interpreter/issues/731 to clarify things up.

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -105,6 +107,63 @@ pub fn random_scalar_in_group_range() -> Scalar {
...
+            for (i, v) in bytes.iter().enumerate() {
+                bytes_32[i] = *v;
+            }
+            Scalar::from_repr(bytes_32.into())
+        }
+        _ => None,
+    }
+}
+
+// /// Attempts to create Scalar from BigInt
```
greenhat, line 143: Looks like we don't need this anymore.

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -105,6 +107,63 @@ pub fn random_scalar_in_group_range() -> Scalar {
     Scalar::generate_vartime(&mut OsRng)
 }
 
+/// Attempts to create BigInt from Scalar
+pub fn scalar_to_bigint(s: Scalar) -> BigInt {
```
greenhat, line 111: A good final touch would be a roundtrip test. No need for PBT, simple one-off `dlog_group::random_scalar_in_group_range()` value converted to BigInt and back.

greenhat (changes requested):
Thank you for your efforts! Just a couple of final touches.

ApexTheory:
Cleaned up the commented code (sorry, late night check-in heh) and added the roundtrip test, definitely a good idea.

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -168,4 +203,16 @@ mod tests {
             prop_assert_eq![sigma_serialize_roundtrip(&e), e];
         }
     }
+
+    #[test]
+    fn scalar_bigint_roundtrip() {
+        let rand_scalar: Scalar = random_scalar_in_group_range();
+        let as_bigint: BigInt = scalar_to_bigint(rand_scalar);
+        let as_scalar: Option<Scalar> = bigint_to_scalar(as_bigint);
```
greenhat, line 211:
```suggestion
        assert_eq!(rand_scalar, bigint_to_scalar(as_bigint).unwrap());
```
This will give the same result without matching.

greenhat (changes requested):
Looking great! This is my last nitpick. :)

ApexTheory:
Updated! Had forgotten about the clippy annotation to allow unwrap.

greenhat (approved):
Great! I appreciate your efforts!

Issue #295, opened 2021-06-08
[250 SigmaUSD] Add 256-bit signed integer type and use it for `Value::BigInt`


greenhat:
### Motivation
In ErgoTree BigInt is bound to be signed 256-bit integer. Rather than using "unbounded" BigInt and check every time that it still fits into 256-bit it's better to make a new type that is guaranteed to fit in 256-bit.

### Implementation
Similar to existing `BoxValue` and `TokenAmount` types create a new type `BigInt256` wrapping `BigInt`.
Use `BigInt256` in `Value::BigInt()`. 
`BigInt256` should implement only one fallible constructor (`TryFrom`) with bounds checking. Arithmetic ops should be failable as well and check the result for bounds. See `Evaluable` impl for `BinOp` for an example of bounds checking.

jerry-best:
I agree this would be a nice change. I had a quick look at it, and saw a couple options -- one is to wrap BigInt as above, and another would be to try using one of the 256-bit int crates available if they look good. I haven't really looked at it, but this one seemed like it could be useful: https://docs.rs/num256/0.3.1/num256/

greenhat:
> I agree this would be a nice change. I had a quick look at it, and saw a couple options -- one is to wrap BigInt as above, and another would be to try using one of the 256-bit int crates available if they look good. I haven't really looked at it, but this one seemed like it could be useful: https://docs.rs/num256/0.3.1/num256/

Good find! Num256 wraps BigInt (https://github.com/althea-net/num256_rs/blob/48350556b930ac71543a8d068cb31923173f5e9b/src/int256.rs#L18), so it might be exactly what we need here.

FitzOReilly:
I've started working on this.

greenhat:
Done in #325 . Bounty is sent.

PR #325, opened 2021-06-27
Implement BigInt256 and use it in Value::BigInt

FitzOReilly:
This is my draft PR for #295. There are still some failing tests after merging the develop branch. I also want to add a few tests for `BigInt256`.

I've created a new type `BigInt256`, which wraps `Int256`, which wraps `BigInt`. I couldn't use `Int256`directly, because it doesn't satisfy all the required traits. I wasn't sure where to put it, so for now it's in `ergotree-ir/src/mir/bigint256.rs`.

In `ergotree-ir/src/sigma_protocol/dlog_group.rs`, I've started to replace the `BigInt`s with `BigInt256`. Is `Scalar` guaranteed to be a 256-bit signed integer? Otherwise, would it make sense to change the return type of `fn scalar_to_bigint(s: Scalar) -> BigInt` to an `Option` type, i.e. `Option<BigInt256>`?


ergotree-ir/Cargo.toml
```diff
@@ -30,6 +30,9 @@ base16 = "0.2.1"
 num-bigint = "0.4.0"
```
greenhat, line 30: I wonder if we still need an explicit dep on `num-bigint` since `num256` pulls it already.

greenhat:
> This is my draft PR for #295. There are still some failing tests after merging the develop branch. I also want to add a few tests for `BigInt256`.

Thank you for taking care of this!
 
> I've created a new type `BigInt256`, which wraps `Int256`, which wraps `BigInt`. I couldn't use `Int256`directly, because it doesn't satisfy all the required traits. I wasn't sure where to put it, so for now it's in `ergotree-ir/src/mir/bigint256.rs`.

Wrapper type sounds good. Please move it to the root (`src`) since `mir` module is for IR nodes. 

> In `ergotree-ir/src/sigma_protocol/dlog_group.rs`, I've started to replace the `BigInt`s with `BigInt256`. Is `Scalar` guaranteed to be a 256-bit signed integer? Otherwise, would it make sense to change the return type of `fn scalar_to_bigint(s: Scalar) -> BigInt` to an `Option` type, i.e. `Option<BigInt256>`?

Yes, `Scalar` is a 256-bit modulo curve order (positive).
Here is how it's defined:
```rust
/// A scalar with arithmetic modulo curve order, represented as 4 64-bit limbs (little-endian).
#[derive(Clone, Copy, Debug, Default)]
#[cfg_attr(docsrs, doc(cfg(feature = "arithmetic")))]
pub struct Scalar4x64([u64; 4]);
```

greenhat:
> > > In `ergotree-ir/src/sigma_protocol/dlog_group.rs`, I've started to replace the `BigInt`s with `BigInt256`. Is `Scalar` guaranteed to be a 256-bit signed integer? Otherwise, would it make sense to change the return type of `fn scalar_to_bigint(s: Scalar) -> BigInt` to an `Option` type, i.e. `Option<BigInt256>`?
> > 
> > 
> > Yes, `Scalar` is a 256-bit modulo curve order (positive).
> > Here is how it's defined:
> > ```rust
> > /// A scalar with arithmetic modulo curve order, represented as 4 64-bit limbs (little-endian).
> > #[derive(Clone, Copy, Debug, Default)]
> > #[cfg_attr(docsrs, doc(cfg(feature = "arithmetic")))]
> > pub struct Scalar4x64([u64; 4]);
> > ```
> 
> Thanks for the definition! I took a look at the `k256` crate.
> But if it's positive, it doesn't fit into a `BigInt256`. So we need to keep the `BigInt` or we would need a `BigUint256`, too. Am I missing something here?

Good catch! Curve order is `FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141` according to https://en.bitcoin.it/wiki/Secp256k1

And in Scala BigInt bounds checked in 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/00ffd5514cb8410c0646f5e10f51e383515fcb67/common/src/main/scala/scalan/util/Extensions.scala#L157-L177
```scala
    /** Checks this {@code BigInteger} can be cust to 256 bit two's-compliment representation,
      * checking for lost information. If the value of this {@code BigInteger}
      * is out of the range of the 256 bits, then an {@code ArithmeticException} is thrown.
      *
      * @return this {@code BigInteger} if the check is successful
      * @throws ArithmeticException if the value of {@code this} will
      * not exactly fit in a 256 bit range.
      * @see BigInteger#longValueExact
      */
    @inline final def to256BitValueExact: BigInteger = {
      // Comparing with 255 is correct because bitLength() method excludes the sign bit.
      // For example, these are the boundary values:
      // (new BigInteger("80" + "00" * 31, 16)).bitLength() = 256
      // (new BigInteger("7F" + "ff" * 31, 16)).bitLength() = 255
      // (new BigInteger("-7F" + "ff" * 31, 16)).bitLength() = 255
      // (new BigInteger("-80" + "00" * 31, 16)).bitLength() = 255
      // (new BigInteger("-80" + "00" * 30 + "01", 16)).bitLength() = 256
      if (x.bitLength() <= 255) x
      else
        throw new ArithmeticException("BigInteger out of 256 bit range");
    }
```
So I guess, yes, `Scalar` can be larger than signed BigInt256. Plus, I suppose that BigInt256 also stores sign in a bit as Scala's version (see code comment above).

FitzOReilly:
> Good catch! Curve order is `FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141` according to https://en.bitcoin.it/wiki/Secp256k1
> 
> And in Scala BigInt bounds checked in
> https://github.com/ScorexFoundation/sigmastate-interpreter/blob/00ffd5514cb8410c0646f5e10f51e383515fcb67/common/src/main/scala/scalan/util/Extensions.scala#L157-L177
> 
> ```scala
>     /** Checks this {@code BigInteger} can be cust to 256 bit two's-compliment representation,
>       * checking for lost information. If the value of this {@code BigInteger}
>       * is out of the range of the 256 bits, then an {@code ArithmeticException} is thrown.
>       *
>       * @return this {@code BigInteger} if the check is successful
>       * @throws ArithmeticException if the value of {@code this} will
>       * not exactly fit in a 256 bit range.
>       * @see BigInteger#longValueExact
>       */
>     @inline final def to256BitValueExact: BigInteger = {
>       // Comparing with 255 is correct because bitLength() method excludes the sign bit.
>       // For example, these are the boundary values:
>       // (new BigInteger("80" + "00" * 31, 16)).bitLength() = 256
>       // (new BigInteger("7F" + "ff" * 31, 16)).bitLength() = 255
>       // (new BigInteger("-7F" + "ff" * 31, 16)).bitLength() = 255
>       // (new BigInteger("-80" + "00" * 31, 16)).bitLength() = 255
>       // (new BigInteger("-80" + "00" * 30 + "01", 16)).bitLength() = 256
>       if (x.bitLength() <= 255) x
>       else
>         throw new ArithmeticException("BigInteger out of 256 bit range");
>     }
> ```
Thanks for guiding me to the right places in the code!
So this just throws an exception if `x > 2^255 - 1`. Then `BigInt256` would be large enough.

> So I guess, yes, `Scalar` can be larger than signed BigInt256. Plus, I suppose that BigInt256 also stores sign in a bit as Scala's version (see code comment above).
Here is `BigInt`s definition:
```rust
/// A big signed integer type.
pub struct BigInt {
    sign: Sign,
    data: BigUint,
}
```
So it stores the sign in a separate variable. So the range could be increased to `-2^256-1 < x < 2^256-1`. I think that would be confusing, though. It also means that `BigInt256` is actually (up to) 256-bit + 1 byte. Maybe this could be refactored in the future to use a fixed size and a two's complement representation.

I feel like I'm getting stuck here.

greenhat:
> > Good catch! Curve order is `FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141` according to https://en.bitcoin.it/wiki/Secp256k1
> > And in Scala BigInt bounds checked in
> > https://github.com/ScorexFoundation/sigmastate-interpreter/blob/00ffd5514cb8410c0646f5e10f51e383515fcb67/common/src/main/scala/scalan/util/Extensions.scala#L157-L177
> > ```scala
> >     /** Checks this {@code BigInteger} can be cust to 256 bit two's-compliment representation,
> >       * checking for lost information. If the value of this {@code BigInteger}
> >       * is out of the range of the 256 bits, then an {@code ArithmeticException} is thrown.
> >       *
> >       * @return this {@code BigInteger} if the check is successful
> >       * @throws ArithmeticException if the value of {@code this} will
> >       * not exactly fit in a 256 bit range.
> >       * @see BigInteger#longValueExact
> >       */
> >     @inline final def to256BitValueExact: BigInteger = {
> >       // Comparing with 255 is correct because bitLength() method excludes the sign bit.
> >       // For example, these are the boundary values:
> >       // (new BigInteger("80" + "00" * 31, 16)).bitLength() = 256
> >       // (new BigInteger("7F" + "ff" * 31, 16)).bitLength() = 255
> >       // (new BigInteger("-7F" + "ff" * 31, 16)).bitLength() = 255
> >       // (new BigInteger("-80" + "00" * 31, 16)).bitLength() = 255
> >       // (new BigInteger("-80" + "00" * 30 + "01", 16)).bitLength() = 256
> >       if (x.bitLength() <= 255) x
> >       else
> >         throw new ArithmeticException("BigInteger out of 256 bit range");
> >     }
> > ```
> 
> Thanks for guiding me to the right places in the code!
> So this just throws an exception if `x > 2^255 - 1`. Then `BigInt256` would be large enough.
> 
> > So I guess, yes, `Scalar` can be larger than signed BigInt256. Plus, I suppose that BigInt256 also stores sign in a bit as Scala's version (see code comment above).
> 
> Here is `BigInt`s definition:
> 
> ```rust
> /// A big signed integer type.
> pub struct BigInt {
>     sign: Sign,
>     data: BigUint,
> }
> ```
> 
> So it stores the sign in a separate variable. So the range could be increased to `-2^256-1 < x < 2^256-1`. I think that would be confusing, though. It also means that `BigInt256` is actually (up to) 256-bit + 1 byte. Maybe this could be refactored in the future to use a fixed size and a two's complement representation.

I'm not sure why you're looking at `BigInt` and not `Int256`. I see the range for `Int256` is `-2^255..=2^255 - 1` here:

https://github.com/althea-net/num256_rs/blob/48350556b930ac71543a8d068cb31923173f5e9b/src/int256.rs#L30-L44

Which looks like exactly what we want.

ergotree-ir/Cargo.toml
FitzOReilly: I tried removing it and got 2 errors: in `bigint256.rs` and in `dlog_group.rs` (which needs `num_bigint::Sign`).

FitzOReilly:
I cleaned this up and think it's ready for review.

I'm still confused by the `Scalar`/`BigInt256` conversion, to be honest, because one of the types is unsigned and the other is signed. So the conversion is fallible. In 2 tests (`scalar_bigint256_roundtrip` in `dlog_group.rs` and `eval_any` in `exponentiate.rs`), I've just shifted the `Scalar` right to make sure it fits into a `BigInt256`. This feels like a dirty hack,but I'm not sure how to handle this in a clean way. If you have any ideas, I'd appreciate it.

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -115,37 +116,30 @@ pub fn random_scalar_in_group_range() -> Scalar {
...
-
-            let mut bytes_32 = [0; 32];
-            for (i, v) in bytes.iter().enumerate() {
-                bytes_32[i] = *v;
-            }
-            Scalar::from_repr(bytes_32.into())
-        }
-        _ => None,
+    let bytes = bytes_be.as_slice();
+    debug_assert!(bytes.len() <= 32);
```
greenhat, line 137: Nice! I've never used `debug_assert!` before. Seems like a good fit here!

ergotree-ir/src/sigma_protocol/dlog_group.rs
```diff
@@ -115,37 +116,30 @@ pub fn random_scalar_in_group_range() -> Scalar {
     Scalar::generate_vartime(&mut OsRng)
 }
 
-/// Attempts to create BigInt from Scalar
-pub fn scalar_to_bigint(s: Scalar) -> BigInt {
+/// Attempts to create BigInt256 from Scalar
+/// Returns None if s > 2^255 - 1
```
greenhat, line 120:
```suggestion
/// Returns None if s > 2^255 - 1
Since Scalar is in [0, n) range, where n is the group order (FFFFFFFF FFFFFFFF FFFFFFFF FFFFFFFE BAAEDCE6 AF48A03B BFD25E8C D0364141) it might not fit into 256-bit BigInt because BigInt uses 1 bit for sign.
```

greenhat (changes requested):
Great job! Please check comments.

greenhat:
> I cleaned this up and think it's ready for review.
> 
> I'm still confused by the `Scalar`/`BigInt256` conversion, to be honest, because one of the types is unsigned and the other is signed. So the conversion is fallible. In 2 tests (`scalar_bigint256_roundtrip` in `dlog_group.rs` and `eval_any` in `exponentiate.rs`), I've just shifted the `Scalar` right to make sure it fits into a `BigInt256`. This feels like a dirty hack,but I'm not sure how to handle this in a clean way. If you have any ideas, I'd appreciate it.

You did a great job! 
I created #348 to dig deeper into the case when Scalar does not fit into BigInt256.

ergotree-ir/src/bigint256.rs
```diff
@@ -0,0 +1,317 @@
...
+impl TryFrom<&[u8]> for BigInt256 {
+    type Error = String;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        let n = BigInt::from_signed_bytes_be(value);
+        Self::try_from(n)
+    }
+}
+
+impl TryFrom<i8> for BigInt256 {
```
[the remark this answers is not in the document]
FitzOReilly: You're right, of course. Changed it (for `i8`, `i16`, `i32` and `i64`) and replaced the `try_from`/`try_into` calls with `from`/`into` where applicable.

ergotree-ir/src/sigma_protocol/dlog_group.rs
FitzOReilly: Added a comment

greenhat (approved):
Looking great! Thanks!

PR #334, opened 2021-07-06
331 - GroupElement - Negate

ApexTheory:
Hi @greenhat, please take a look at your convenience.

*Edit - not sure why the ubuntu-latest build is failing on the scalar/bigint roundtrip test in dlog_group.  
- This code was not touched and it passed in the previous PR.  
- I do see that this build is using a slightly newer virtual environment (20210628.1 vs 20210621.1)
- It is working fine on my local (Windows) machine. 
- It passed in my [new PR](https://github.com/ergoplatform/sigma-rust/pull/335) for GroupElement.getEncoded **which contains these commits**

image

PR #335, opened 2021-07-08
330 - GroupElement - getEncoded

ApexTheory:
Hi @greenhat, please review at your convenience.  

Please note, this branch is based off my [previous negate branch](https://github.com/ergoplatform/sigma-rust/pull/334) (as it requires some common code), so it should be merged afterwards.

### DIFF
--- a/ergotree-ir/src/types/sgroup_elem.rs
+++ b/ergotree-ir/src/types/sgroup_elem.rs
@@ -7,7 +7,6 @@ use super::smethod::SMethodDesc;
 use super::stype::SType;
 use super::stype_companion::STypeCompanion;
 use super::stype_companion::STypeCompanionHead;
-use super::stype_param::STypeVar;
 use lazy_static::lazy_static;
 
 /// SGroupElement type id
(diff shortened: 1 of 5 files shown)


ergotree-ir/src/types/sgroup_elem.rs
```diff
@@ -0,0 +1,73 @@
...
+    );
+}
+
+lazy_static! {
+    static ref GET_ENCODED_METHOD_DESC: SMethodDesc = SMethodDesc {
+        method_id: GET_ENCODED_METHOD_ID,
+        name: "getEncoded",
+        tpe: SFunc::new(
+            vec![SType::SGroupElement],
+            SType::SColl(SType::STypeVar(STypeVar::ov()).into()),
```
greenhat, line 42: The result type should be `Coll[Byte]`.

greenhat (changes requested):
Looking great! Only one comment regarding result type.

ergotree-ir/src/types/sgroup_elem.rs
greenhat: `STypeVar` is used when we need to express a polymorphic type (think `Coll.map`, etc.).

ApexTheory: Fixed.

greenhat (approved):
Great! Thanks!

PR #340, opened 2021-07-09
332 - GlobalVar - GroupGenerator

### LINKED ISSUE #332: [200 SigmaUSD] Implement `groupGenerator` global var
- Add enum variant `GroupGenerator` to `ergotree-ir::mir::global_vars::GlobalVars`;
- Write serialization roundtrip tests for all `GlobalVars` variants (PBT will be suffice, see `serialization::expr::tests::ser_roundtrip`);
- Add `GlobalGenerator` enum variant handling in `ergotree-interpreter::eval::global_vars`;

Reference Scala version:
IR node and eval - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/b60edae20f2af3352e4064d139d45a15fc507f7d/sigmastate/src/main/scala/sigmastate/Values.scala#L691-L701

ApexTheory:
GroupGenerator added to global vars.  I believe I have this correct, to use the dlog_group::generator?

EDIT(@greenhat ): Close #332

### DIFF
--- a/ergotree-interpreter/src/eval/global_vars.rs
+++ b/ergotree-interpreter/src/eval/global_vars.rs
@@ -1,11 +1,12 @@
 use crate::eval::Env;
 use ergotree_ir::mir::global_vars::GlobalVars;
 use ergotree_ir::mir::value::Value;
+use ergotree_ir::serialization::SigmaSerializable;
+use ergotree_ir::sigma_protocol::dlog_group;
 
 use super::EvalContext;
 use super::EvalError;
 use super::Evaluable;
-use ergotree_ir::serialization::SigmaSerializable;
 
 impl Evaluable for GlobalVars {
     fn eval(&self, _env: &Env, ectx: &mut EvalContext) -> Result<Value, EvalError> {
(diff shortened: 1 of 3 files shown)


ergotree-interpreter/src/eval/global_vars.rs
```diff
@@ -68,4 +71,11 @@ mod tests {
             ctx.inputs
         );
     }
+
+    #[test]
+    fn eval_group_generator() {
+        let ctx = Rc::new(force_any_val::<Context>());
+        let res = eval_out::<EcPoint>(&GlobalVars::GroupGenerator.into(), ctx.clone());
+        assert!(matches!(res.into(), Value::GroupElement(_)));
```
greenhat, line 79: For regression test checking that evaluation returns an exact value (`dlog_group::generator()`) would be a better fit.

greenhat (changes requested):
Looking good! Thanks! Just one comment about the test.

ergotree-interpreter/src/eval/global_vars.rs
ApexTheory: Fixed.

greenhat (approved):
Thanks!

PR #341, opened 2021-07-12
338 - SigmaBoolean serialization

### LINKED ISSUE #338: Serialization of `SigmaBoolean`
Implement serialization and parsing for all missing `SigmaBoolean` variants (see `todo!()` - `Cand`, `Cor`, parsing for `ProveDhTuple`). 

https://github.com/ergoplatform/sigma-rust/blob/6698880353fbec1038d0947ca2488746bcb77f34/ergotree-ir/src/serialization/sigmaboolean.rs#L14-L46

Write roundtrip serialization test (implement `Arbitrary` for `SigmaBoolean` first).

Reference Scala implementation - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/00ffd5514cb8410c0646f5e10f51e383515fcb67/sigmastate/src/main/scala/sigmastate/Values.scala#L610-L684

ApexTheory:
Hey @greenhat please take a look at your convenience.  I used the existing rust & scala code as a reference to fill in the gaps.  

I do have some questions regarding the Arbitrary impl and the roundtrip test, I will add them alongside the relevant code in the PR.

EDIT(@greenhat): Close #338

### DIFF
--- a/ergotree-ir/src/sigma_protocol/sigma_boolean.rs
+++ b/ergotree-ir/src/sigma_protocol/sigma_boolean.rs
@@ -69,6 +69,10 @@ pub struct ProveDhTuple {
     pub v: Box<EcPoint>,
 }
 
+impl HasStaticOpCode for ProveDhTuple {
+    const OP_CODE: OpCode = OpCode::PROVE_DIFFIE_HELLMAN_TUPLE;
+}
+
 impl ProveDhTuple {
     /// Create new instance
     pub fn new(g: EcPoint, h: EcPoint, u: EcPoint, v: EcPoint) -> Self {
@@ -79,10 +83,6 @@ impl ProveDhTuple {
             v: v.into(),
         }
     }
-}
-
-impl HasStaticOpCode for ProveDhTuple {
-    const OP_CODE: OpCode = OpCode::PROVE_DIFFIE_HELLMAN_TUPLE;
 }
 
 /// Sigma proposition
@@ -173,9 +173,67 @@ impl From<ProveDlog> for SigmaBoolean {
     }
 }
 
+impl TryInto<ProveDhTuple> for SigmaBoolean {
+    type Error = ConversionError;
+    fn try_into(self) -> Result<ProveDhTuple, Self::Error> {
+        match self {
+            SigmaBoolean::ProofOfKnowledge(SigmaProofOfKnowledgeTree::ProveDhTuple(pdh)) => Ok(pdh),
+            _ => Err(ConversionError),
+        }
+    }
+}
+
 impl From<ProveDhTuple> for SigmaBoolean {
     fn from(v: ProveDhTuple) -> Self {
         SigmaBoolean::ProofOfKnowledge(SigmaProofOfKnowledgeTree::ProveDhTuple(v))
+    }
+}
+
+impl TryInto<Cand> for SigmaBoolean {
+    type Error = ConversionError;
+    fn try_into(self) -> Result<Cand, Self::Error> {
+        match self {
+            SigmaBoolean::SigmaConjecture(SigmaConjecture::Cand(c)) => Ok(c),
+            _ => Err(ConversionError),
+        }
+    }
+}
+
+impl From<Cand> for SigmaBoolean {
+    fn from(v: Cand) -> Self {
+        SigmaBoolean::SigmaConjecture(SigmaConjecture::Cand(v))
+    }
+}
+
+impl TryInto<Cor> for SigmaBoolean {
+    type Error = ConversionError;
+    fn try_into(self) -> Result<Cor, Self::Error> {
+        match self {
+            SigmaBoolean::SigmaConjecture(SigmaConjecture::Cor(c)) => Ok(c),
+            _ => Err(ConversionError),
+        }
+    }
+}
+
+impl From<Cor> for SigmaBoolean {
+    fn from(v: Cor) -> Self {
+        SigmaBoolean::SigmaConjecture(SigmaConjecture::Cor(v))
+    }
+}
+
+impl TryInto<Cthreshold> for SigmaBoolean {
+    type Error = ConversionError;
+    fn try_into(self) -> Result<Cthreshold, Self::Error> {
+        match self {
+            SigmaBoolean::SigmaConjecture(SigmaConjecture::Cthreshold(c)) => Ok(c),
+            _ => Err(ConversionError),
+        }
+    }
+}
+
+impl From<Cthreshold> for SigmaBoolean {
+    fn from(v: Cthreshold) -> Self {
+        SigmaBoolean::SigmaConjecture(SigmaConjecture::Cthreshold(v))
     }
 }
 
@@ -230,8 +288,10 @@ impl From<ProveDhTuple> for SigmaProp {
 }
 /// Arbitrary impl for ProveDlog
 #[cfg(feature = "arbitrary")]
+#[allow(clippy::unwrap_used)]
 mod arbitrary {
     use super::*;
+    use proptest::collection::vec;
     use proptest::prelude::*;
 
     impl Arbitrary for ProveDlog {
@@ -243,14 +303,49 @@ mod arbitrary {
         }
     }
 
-    impl Arbitrary for SigmaBoolean {
+    impl Arbitrary for ProveDhTuple {
         type Parameters = ();
         type Strategy = BoxedStrategy<Self>;
 
         fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-            (any::<ProveDlog>())
-                .prop_map(|p| {
-                    SigmaBoolean::ProofOfKnowledge(SigmaProofOfKnowledgeTree::ProveDlog(p))
+            (
+                any::<EcPoint>(),
+                any::<EcPoint>(),
+                any::<EcPoint>(),
+                any::<EcPoint>(),
+            )
+                .prop_map(|(g, h, u, v)| ProveDhTuple::new(g, h, u, v))
+                .boxed()
+        }
+    }
+
+    pub fn primitive_type_value() -> BoxedStrategy<SigmaBoolean> {
+        prop_oneof![
+            any::<ProveDlog>().prop_map_into(),
+            any::<ProveDhTuple>().prop_map_into(),
+        ]
+        .boxed()
+    }
+
+    impl Arbitrary for SigmaBoolean {
+        type Parameters = ();
+        type Strategy = BoxedStrategy<Self>;
+
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            primitive_type_value()
+                .prop_recursive(1, 8, 4, |elem| {
+                    prop_oneof![
+                        vec(elem.clone(), 2..=4)
+                            .prop_map(|elems| Cand {
+                                items: elems.try_into().unwrap()
+                            })
+                            .prop_map_into(),
+                        vec(elem, 2..=4)
+                            .prop_map(|elems| Cor {
+                                items: elems.try_into().unwrap()
+                            })
+                            .prop_map_into(),
+                    ]
                 })
                 .boxed()
         }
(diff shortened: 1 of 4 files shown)


ergotree-ir/src/sigma_protocol/sigma_boolean.rs
```diff
@@ -196,6 +211,22 @@ mod arbitrary {
...
+                any::<EcPoint>(),
+                any::<EcPoint>(),
+                any::<EcPoint>(),
+            )
+                .prop_map(|(gv, hv, uv, vv)| ProveDhTuple::new(gv, hv, uv, vv))
+                .boxed()
+        }
+    }
+
     impl Arbitrary for SigmaBoolean {
```
ApexTheory, line 230: There was an existing SigmaBoolean Arbitrary impl here that uses the ProveDlog enum variant.  
Do we implementations for the other variants (ProveDHTuple, Cand, Cor, Cthreshold)

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
```diff
@@ -220,4 +251,16 @@ mod arbitrary {
...
+mod tests {
+    use super::*;
+    use crate::serialization::sigma_serialize_roundtrip;
+    use proptest::prelude::*;
+
+    proptest! {
+
+        #[test]
+        fn sigma_boolean_ser_roundtrip(v in any::<SigmaBoolean>()) {
+            prop_assert_eq![sigma_serialize_roundtrip(&v), v]
```
ApexTheory, line 263: When I print out the generated SigmaBoolean value on my local machine, I ALWAYS get the following:
ProofOfKnowledge(ProveDlog(ProveDlog { h: EC:0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798 }))

Should this value not be changing as it is randomly generated?

greenhat: Good catch! I've been thinking about that for a while. Although we have `DlogProverInput::random()` from which we can get a random `EcPoint` with `DlogProverInput::public_image()` it's in `ergotree-interpreter` crate. I was thinking of moving it to `ergotree-ir` but eventually declined since it belongs to an "interpreter's" part of the sigma protocol.
So I suggest to make `EcPoint::random()` (exp the group generator) and use it in `Arbitrary` impl for `EcPoint` (see `Arbitrary` impl for `DlogProverInput`).

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
greenhat: Good question! Please, extend `SigmaBoolean` impl with DH tuple `Cand` and `Cor` conjectures and adding `Arbitrary` impl for them along the way. 
Be aware that `SigmaBoolean` is extensively used in PBT tests (through `SigmaProp`) for boxes, transaction building and signing, etc. Watch for the impact on overall test suite run time. We need to keep it reasonable. :)

greenhat (commented):
Looking good! Thanks for taking care of this!

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
ApexTheory: Would something like this work?   
`    
        fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
            prop_oneof![
                (any::<ProveDlog>()),
                (any::<ProveDhTuple>()),
                (any::<Cand>()),
                (any::<Cor>()),
                (any::<Cthreshold>()),
            ]
            .prop_map(|p| match p {
                SigmaProofOfKnowledgeTree::ProveDlog(provedlog) => {
                    SigmaBoolean::ProofOfKnowledge(SigmaProofOfKnowledgeTree::ProveDlog(provedlog))
                }
                SigmaProofOfKnowledgeTree::ProveDhTuple(provedhtuple) => {
                    SigmaBoolean::ProofOfKnowledge(SigmaProofOfKnowledgeTree::ProveDhTuple(
                        provedhtuple,
                    ))
                }
                SigmaConjecture::Cand(cand) => {
                    SigmaBoolean::SigmaConjecture(SigmaConjecture::Cand(cand))
                }
                SigmaConjecture::Cor(cor) => {
                    SigmaBoolean::SigmaConjecture(SigmaConjecture::Cor(cor))
                }
                SigmaConjecture::Cthreshold(cthreshold) => {
                    SigmaBoolean::SigmaConjecture(SigmaConjecture::Cthreshold(cthreshold))
                }
            })
            .boxed()
        }
    }
`

Also, I am still working on the `Arbitrary` impl for `Cand`/`Cor`...it seems I need to bound the `SigmaConjectureItems` vectors between 2 and 1000?  Maybe its better to keep the size small to minimize test time as you mentioned?

greenhat: 1. There is a `prop_recursive` for the generation of recursive structures. See how it's used in constant generator - https://github.com/ergoplatform/sigma-rust/blob/e4965fcd758c0d154bb81ec424d4c70a189a8514/ergotree-ir/src/mir/constant.rs#L313-L359 . 
2. There should be `From` implemented for `SigmaBoolean` so that would allow simple `.into()` or `.prop_map_into()`.
3. Yes, something like `2..5` should suffice for conjecture items.

ApexTheory:
Hi @greenhat , I pushed a lot of changes into the branch; I am now able to generate the arbitrary `SigmaBoolean` values (either a proof or conjecture based on param), but there are a few things I likely need to look at further:
1) prop_recursive params - I just defaulted everything to 1 for now and did not explore much beyond
2) should the `SigmaBoolean ` Arbitrary impl be calling the `Cand`/`Cor` Arbitary impl?
3) is use of unwrap okay within the `Cand`/`Cor` From impl?

Please take a look when you have some time, thanks!

ergotree-ir/src/sigma_protocol/sigma_boolean/cand.rs
```diff
@@ -45,11 +48,55 @@ impl Cand {
...
+        self.items.sigma_serialize(w)
+    }
+
+    fn sigma_parse<R: SigmaByteRead>(r: &mut R) -> Result<Self, SigmaParsingError> {
+        let items = SigmaConjectureItems::<_>::sigma_parse(r)?;
+        Ok(Cand { items })
+    }
+}
+
+impl From<Vec<SigmaBoolean>> for Cand {
```
greenhat, line 62: We should avoid using unwrap in production code. It's perfectly fine in tests though. 
`From` is not applicable here, since we want the error on an invalid number of items to be handled by the caller. `TryFrom` fits here, but be aware that `BoundedVec` (type behind a `SigmaConjectureItems` type alias) has one.

greenhat (changes requested):
Looking good!  Thank you!

> Hi @greenhat, I pushed a lot of changes into the branch (and rebased on develop); I am now able to generate the arbitrary `SigmaBoolean` values (either a proof or conjecture based on param), but there are a few things I likely need to look at further:
> 
> 1. `prop_recursive `params - I just defaulted everything to 1 for now and did not explore much beyond

I'd say this might "choke" `prop_recursive`. I suggested 16 for total element target (think of it as total nodes+leafs count of the resulted tree). `expected_branch_size` is a bit of a mystery to me, but setting it to the half of the `desired_size` seems like does the job fine. :)

> 2. ~should the SigmaBoolean Arbitrary impl be calling the Cand/Cor Arbitary impl?~ I think not

Yes, "inside" `prop_recursive`. I don't see any reason not to.

> 3. is use of unwrap okay within the `Cand`/`Cor` From impl?

No. See the reasoning in the comment.

> 4. calling `any_with::<SigmaBoolean>(ArbSigmaBooleanParams::SigmaConjecture(1))` seems to return ProofOfKnowledge values along with Conjectures, when I expect only Conjectures.  Perhaps some mechanism of `prop_oneof!` I am not understanding.

I believe it's `prop_recursive` feature to generate trees that only consist of one leaf.

> 5. impact on overall test run time

We should not significantly slow down the tests. Ideally, it should add some 5%-10%. If it gets bad, decrease `prop_recursive` parameters further.

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
```diff
@@ -223,19 +300,72 @@ mod arbitrary {
...
+    }
+
+    impl Arbitrary for SigmaBoolean {
+        type Parameters = ArbSigmaBooleanParams;
+        type Strategy = BoxedStrategy<Self>;
+
+        fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+            match args {
+                ArbSigmaBooleanParams::ProofOfKnowledge => primitive_type_value(),
+                ArbSigmaBooleanParams::SigmaConjecture(depth) => primitive_type_value()
```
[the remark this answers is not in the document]
ApexTheory: Yes this is a good point.  I thought there may be some benefit to request only primitives via parameter.  Will update.

ergotree-ir/src/sigma_protocol/sigma_boolean/cand.rs
ApexTheory: Makes sense.  So I am removing this code from `Cand`/`Cor`, and I am okay to unwrap() in the `SigmaBoolean` Arbitrary impl as below (since the caller is test code and does not need to handle the error gracefully). 
```
Cand { 
   items: BoundedVec::try_from(items).unwrap() 
}
```

Will check in both changes.

ApexTheory:
I pushed some more changes including the following:

1. updating the prop_recursive params to your suggested defaults
2. call `Cand/Cor::any` from within `SigmaBoolean`'s prop_recursive
3. removed `Cand/Cor` From impl and instead using `BoundedVec::try_from`
4. only using unwrap() in test code
5. removed `ArbSigmaBooleanParams`

One issue is that I am not able to generate nested conjectures using `any::<SigmaBoolean>()`.  It seems to be always a single conjecture with several proofs.  Need to explore the recursive parameters further.

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
```diff
@@ -316,53 +315,24 @@ mod arbitrary {
...
-                                )))
-                            }),
-                        ]
-                    })
-                    .boxed(),
-            }
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            primitive_type_value()
+                .prop_recursive(2, 16, 8, |_elem| {
+                    prop_oneof![any::<Cand>().prop_map_into(), any::<Cor>().prop_map_into(),]
```
greenhat, line 333: You don't need to call `any<Cand>()` here. `_elem` is a generator (`Strategy`) which should be used to produce a few values (via `prop::collection::vec(elem, 2..5)`) with something like this:
```rust
prop_oneof![
    prop::collection::vec(elem, 2..=4)
        .prop_map(|elems| Cand { items: elems.try_into().unwrap() }),
    prop::collection::vec(elem, 2..=4)
        .prop_map(|elems| Cor { items: elems.try_into().unwrap() })
]
```

By calling `any<Cand>()` here you're ignoring "recursive" part of `prop_recursive` producing "generic" `Cor` and `Cand`. 

This is also the reason why you cannot call `any<SigmaBoolean>()` in `Arbitrary` impl for `Cor`/`Cand`. Because in the end, it tries to call "itself".

ergotree-ir/src/sigma_protocol/sigma_boolean/cand.rs
```diff
@@ -59,33 +60,26 @@ impl SigmaSerializable for Cand {
...
         type Strategy = BoxedStrategy<Self>;
 
         fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-            prop_oneof![vec(
-                any_with::<SigmaBoolean>(ArbSigmaBooleanParams::ProofOfKnowledge),
-                2..5,
-            )
-            .prop_map_into(),]
-            .boxed()
+            vec(primitive_type_value(), 2..5)
```
greenhat, line 78: After the fix above you should be able to call `any<SigmaBoolean>()` here.

greenhat (changes requested):
Looking good! We're getting there, `prop_recursive` is hard! :)

ergotree-ir/src/sigma_protocol/sigma_boolean.rs
ApexTheory: Thanks for the feedback, this makes sense.  I have it working with your suggestions, will push the changes now.

greenhat (changes requested):
Looking great! I see tests are taking around 50% longer to run (comparing CI ubuntu-latest with develop). That’s a bit too much. Please try cutting down prop_recursive parameters to `2, 8, 4` and if it does not help even to `1, 8, 4`. Let’s try to get it down to 20%-30% slowdown tops.

greenhat (approved):

Issue #356, opened 2021-08-02
[150 SigmaUSD] Implement `XorOf` IR node (XOR for collection of booleans)


greenhat:
Check out similar `And` IR node - https://github.com/ergoplatform/sigma-rust/blob/4784e17cf026844d40536c1da4ac20231de1ad55/ergotree-ir/src/mir/and.rs#L14-L20

Create a new struct in a separate module(file) in `mir` module and add it to `Expr` enum. 

### Serialization
- Implement `Arbitrary` trait and add serialization roundtrip PBT test for this node;

### Evaluation
- Implement `Evaluable` trait in `ergotree-interpreter` crate;
- Add evaluation tests;

Check out the detailed implementation instructions for `Append`(`Coll`) op in  #301

Scala reference implementation:
IR node - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/0c5c5b515ccb452536e440469b56d763d5ba30ec/sigmastate/src/main/scala/sigmastate/trees.scala#L348-L375
Serialization - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/0c5c5b515ccb452536e440469b56d763d5ba30ec/sigmastate/src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala#L11-L22 (the same as for `And` IR node).

ApexTheory:
PR open https://github.com/ergoplatform/sigma-rust/pull/379

greenhat:
Done in #379. Bounty is sent.

PR #379, opened 2021-08-19
XorOf IR Node implementation

ApexTheory:
PR for https://github.com/ergoplatform/sigma-rust/issues/356

Since `Xor` is associative, I went with a slightly different approach than the Scala implementation based on https://github.com/ScorexFoundation/sigmastate-interpreter/issues/640 by checking for an odd number of `true` values.

Let me know if you see any issues with this; instead I can rework this to use `^` bitwise XOR, perhaps in a `fold` or `reduce` operation.

In the scala version, there is also a check for `E.context.activatedScriptVersion >= 2`, do I need to take this into account?

### DIFF
--- a/ergotree-interpreter/src/eval/xor_of.rs
+++ b/ergotree-interpreter/src/eval/xor_of.rs
@@ -1,0 +1,40 @@
+use ergotree_ir::mir::constant::TryExtractInto;
+use ergotree_ir::mir::value::Value;
+
+use crate::eval::env::Env;
+use crate::eval::EvalContext;
+use crate::eval::EvalError;
+use crate::eval::Evaluable;
+use ergotree_ir::mir::xor_of::XorOf;
+
+impl Evaluable for XorOf {
+    fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let input_v = self.input.eval(env, ctx)?;
+        let input_v_bools = input_v.try_extract_into::<Vec<bool>>()?;
+        Ok((input_v_bools.into_iter().filter(|x| *x).count() & 1 == 1).into())
+    }
+}
+
+#[allow(clippy::panic)]
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::eval::context::Context;
+    use crate::eval::tests::eval_out;
+    use ergotree_ir::mir::expr::Expr;
+    use proptest::collection;
+    use proptest::prelude::*;
+    use sigma_test_util::force_any_val;
+    use std::rc::Rc;
+
+    proptest! {
+
+        #[test]
+        fn eval(bools in collection::vec(any::<bool>(), 10..=20)) {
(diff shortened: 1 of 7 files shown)


ergotree-interpreter/src/eval/xor_of.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::eval::EvalContext;
+use crate::eval::EvalError;
+use crate::eval::Evaluable;
+use ergotree_ir::mir::xor_of::XorOf;
+
+impl Evaluable for XorOf {
+    fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let input_v = self.input.eval(env, ctx)?;
+        let input_v_bools = input_v.try_extract_into::<Vec<bool>>()?;
+        Ok((input_v_bools.into_iter().filter(|x| *x).count() & 1 == 1).into())
```
greenhat, line 14: Frankly, I don't find it very readable. Although I'm biased here, since I was the one who botched the Scala implementation - https://github.com/ScorexFoundation/sigmastate-interpreter/issues/640 and now I'm blowing on the water. So let's leave it as is. :)

ergotree-interpreter/src/eval/xor_of.rs
```diff
@@ -0,0 +1,40 @@
...
+    use std::rc::Rc;
+
+    proptest! {
+
+        #[test]
+        fn eval(bools in collection::vec(any::<bool>(), 10..=20)) {
+            let expr: Expr = XorOf {input: Expr::Const(bools.clone().into()).into()}.into();
+            let ctx = Rc::new(force_any_val::<Context>());
+            let res = eval_out::<bool>(&expr, ctx);
+            prop_assert_eq!(res, bools.into_iter().filter(|x| *x).count() & 1 == 1);
```
greenhat, line 37: There is not much sense in testing it's semantic equivalence with itself. Let's use a different algo here, whatever you prefer. Also, please test on empty collection as well (extend range to 0..=20).

greenhat (changes requested):
Looks good! Thank you! Please see a comment about testing.

You can ignore the check in Scala version.

ergotree-interpreter/src/eval/xor_of.rs
ApexTheory: I updated the test with new range (forgot to revert after doing some personal testing) and also to use a mutable val to hold the reassigned result after each XOR gate.  To your point, this solution is likely more readable than the eval function, so if you prefer maybe I swap them?

greenhat: Yes, please swap it. I find XOR gate with mutable accum more readable.

ApexTheory:
I swapped the algos between eval and test.  Eval now uses an accumulator via Iterator's `fold` method.

greenhat (approved):
Great! Fold looks even better here!

Issue #360, opened 2021-08-02
[250 SigmaUSD] Implement `Option.map`, `filter` methods


greenhat:
Create `ergotree_ir::types::soption` module, add `STypeCompanion` and `SMethod` instances (see `sgroup_elem` module for examples).
Add `EvalFn` in `ertotree-interpreter::eval::soption` and include it in `ergotree-interpreter::eval::smethod_eval_fn`

See existing method implementations in `sgroup_elem`, `scoll` and `sbox` for examples;

Reference Scala version:
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/0c5c5b515ccb452536e440469b56d763d5ba30ec/sigmastate/src/main/scala/sigmastate/types.scala#L1283-L1298

greenhat:
Done in #388.

greenhat:
Bounty is sent.

PR #388, opened 2021-09-04
Option map & filter implementation

ApexTheory:
PR for https://github.com/ergoplatform/sigma-rust/issues/360

### DIFF
--- a/ergotree-interpreter/src/eval/soption.rs
+++ b/ergotree-interpreter/src/eval/soption.rs
@@ -69,22 +69,27 @@ pub(crate) static FILTER_EVAL_FN: EvalFn = |env, ctx, obj, args| {
         ))),
     }?;
 
-    if let Some(v) = normalized_input_val {
-        if let Value::Boolean(predicate_res) = predicate_call(v.clone())? {
-            if predicate_res {
-                return Ok(Value::Opt(Box::new(Some(v))));
-            }
-        }
+    match normalized_input_val {
+        Some(val) => match predicate_call(val.clone())? {
+            Value::Boolean(predicate_res) => match predicate_res {
+                true => Ok(Value::Opt(Box::new(Some(val)))),
+                false => Ok(Value::Opt(Box::new(None))),
+            },
+            _ => Err(EvalError::UnexpectedValue(format!(
+                "expected filter predicate result to be boolean, got: {0:?}",
+                lambda.body.tpe()
+            ))),
+        },
+        None => Ok(Value::Opt(Box::new(None))),
     }
-    Ok(Value::Opt(Box::new(None)))
 };
 
 #[allow(clippy::unwrap_used)]
 #[cfg(test)]
 #[cfg(feature = "arbitrary")]
 mod tests {
-    use ergotree_ir::mir::bin_op::BinOp;
     use ergotree_ir::mir::bin_op::RelationOp;
+    use ergotree_ir::mir::bin_op::{ArithOp, BinOp};
     use ergotree_ir::mir::constant::Constant;
     use ergotree_ir::mir::expr::Expr;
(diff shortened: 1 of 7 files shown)


ergotree-ir/src/mir/value.rs
```diff
@@ -432,7 +432,7 @@ impl<T: TryExtractFrom<Value>> TryExtractFrom<Value> for Option<T> {
         match v {
             Value::Opt(opt) => opt.map(T::try_extract_from).transpose(),
             _ => Err(TryExtractFromError(format!(
-                "expected Context, found {:?}",
+                "expected Option, found {:?}",
```
greenhat, line 435: Thanks for the fix!

ergotree-ir/src/types/soption.rs
```diff
@@ -0,0 +1,83 @@
...
+        name: "filter",
+        tpe: SFunc::new(
+            vec![
+                SType::SOption(SType::STypeVar(STypeVar::iv()).into()),
+                SFunc::new(
+                    vec![STypeVar::iv().into()],
+                    SType::SBoolean,
+                ).into()
+                ],
+            SType::SOption(SType::STypeVar(STypeVar::ov()).into()),
```
greenhat, line 67:
```suggestion
            SType::SOption(SType::STypeVar(STypeVar::iv()).into()),
```

ergotree-interpreter/src/eval/soption.rs
```diff
@@ -0,0 +1,325 @@
...
+    let normalized_input_val: Option<Value> = match input_v {
+        Value::Opt(opt) => Ok(*opt),
+        _ => Err(EvalError::UnexpectedValue(format!(
+            "expected filter input to be Value::Opt, got: {0:?}",
+            input_v
+        ))),
+    }?;
+
+    if let Some(v) = normalized_input_val {
+        if let Value::Boolean(predicate_res) = predicate_call(v.clone())? {
```
greenhat, line 73: I believe if `predicate_call` return something other than `Value::Boolean` we silently return `None`. I'd rather return an error in this case.

greenhat (approved):
Looking great overall! I appreciate the test coverage! Please see my comments.

ergotree-interpreter/src/eval/soption.rs
ApexTheory: So I figured that this case would be handled by `MethodCall::new` where it checks expected types according to the `SMethodDesc`.  In a unit test, I am unable to even instantiate the `MethodCall` if the predicate's `Expr.tpe()` is not `SType::SBoolean`.

In any case, I have added an explicit check for this in the eval function as well.  I switched to match statements to make it more readable.

ergotree-ir/src/types/soption.rs
ApexTheory: Thanks, good catch.

ergotree-interpreter/src/eval/soption.rs
greenhat: Type check on IR node construction checks the declared type while on the evaluation we have a resulted value that is not guaranteed to have that declared type. Think of not-yet-evaluated expression as an opaque box with a sticker(type) on it. Only when it is opened (expression is evaluated) we can see the result value and its type. Due to possible errors on evaluation, compilation, serialization, etc. I prefer to "double" check that what we got is the same that what was expected (sticker) before propagating an invalid value further up the expression tree.

greenhat (approved):
Looking great!

ergotree-interpreter/src/eval/soption.rs
ApexTheory: Makes sense, thanks for the explanation.

PR #391, opened 2021-09-08
Rename UnaryOp* traits to OneArgOp*

### LINKED ISSUE #293: Rename `UnaryOp*` traits to `OneArgOp*`
### Motivation
When introducing similar traits for IR nodes that have two arguments one would expect to name them `BinaryOp*`, but we already have `BinOp` that serves a completely different purpose. Instead, the names `TwoArgOp*` were used in #292. So, in order to be consistent, I suggest we rename `UnaryOp*` traits to `OneArgOp*`.

kettlebell:
Close #293


greenhat (approved):
Thanks for taking care of this!

PR #394, opened 2021-09-10
Simplify more *Error conversions in ergo-lib-wasm

### LINKED ISSUE #97: Simplify *Error to JsValue conversion
The goal is to avoid `.map_err(|e| JsValue::from_str(&format!("{}", e)))` in WASM layer. Ideally, I see implementing `From/Into` for errors and use `?` but orphan rules should not allow that.

See code example - https://github.com/ergoplatform/sigma-rust/blob/develop/bindings/ergo-lib-wasm/src/contract.rs#L11-L20

kettlebell:
Covers all remaining cases. Close #97

### DIFF
--- a/bindings/ergo-lib-wasm/Cargo.toml
+++ b/bindings/ergo-lib-wasm/Cargo.toml
@@ -14,6 +14,7 @@ crate-type = ["cdylib", "rlib"]
 default = ["console_error_panic_hook"]
 
 [dependencies]
+base16 = "0.2.1"
 serde = { version = "1.0", features = ["derive"]}
 ergo-lib = { version = "^0.13.0", path = "../../ergo-lib" }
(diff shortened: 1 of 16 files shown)


greenhat (approved):
Looking great!

What do you think of renaming `conv` to `to_js`? It just came to my mind that `conv` sounds a bit too abstract for our use case. And `map_err(to_js)` looks more readable/comprehensible to newcomers.

kettlebell:
> Looking great!
> 
> What do you think of renaming `conv` to `to_js`? It just came to my mind that `conv` sounds a bit too abstract for our use case. And `map_err(to_js)` looks more readable/comprehensible to newcomers.

`to_js` is much better. Will change.

greenhat (approved):
Great! Thanks!

PR #395, opened 2021-09-12
Change constant v type

kettlebell:
For issue #343.

In summary, I've split out a `Literal` type to replace `Value` in `Constant::v`. Also `DataSerializer` was changed to serialise/parse `Literal` instead of `Value`.

PR #396, opened 2021-09-15
Change `Constant::v` type

### LINKED ISSUE #343: Consider switching `Constant::v` type from `Value`
### Motivation
- Some of the variants of `Value` cannot be serialized (see https://github.com/ergoplatform/sigma-rust/blob/dceb878840a57ac25695818a2137a04e2be21535/ergotree-ir/src/serialization/data.rs#L74-L79
), thus hinting that separate enum for constant values (`Literal`?) might be a better fit. Or even "embedded" enum for such values in `Constant`. After all, `Value` represents "runtime" values during the interpreter run and might not be suited to represent literal values in code.
- `Value::CBox` holds a reference (`IrBoxId`) to the box stored in the interpreter's context (`Context`) which is unavailable during the serialization.

kettlebell:
For issue #343. A corrected PR from the aborted PR #395.

EDIT(@greenhat ): Close #343

### DIFF
--- a/ergotree-interpreter/src/eval/expr.rs
+++ b/ergotree-interpreter/src/eval/expr.rs
@@ -10,7 +10,7 @@ impl Evaluable for Expr {
     fn eval(&self, env: &Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
         ctx.cost_accum.add_cost_of(self)?;
         match self {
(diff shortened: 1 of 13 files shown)


greenhat (approved):
Looking great! Thank you!

PR #398, opened 2021-09-16
Add field aliases for Explorer API

### LINKED ISSUE #152: Ensure JSON from explorer API is parsed properly
### Motivation
Node and explorer use slightly different names for the same fields. For example `id` vs `boxId` for box ids, `transactionId` vs `txId` for transaction id in box, which are handled as aliases in serde.

### Fix
Find all field names that have the same meaning but different names in explorer vs node API. Use serde alias to handle different names for the same field. See [example](https://github.com/ergoplatform/sigma-rust/blob/json-error-transactionid/ergo-lib/src/chain/json.rs#L74)
One way to find the differences is to compare `openapi.yml`.
Node: https://github.com/ergoplatform/ergo/blob/develop/src/main/resources/api/openapi.yaml
Explorer: https://api.ergoplatform.com/docs/openapi

kettlebell:
Note: v1 of the Explorer API is consistent with the Node API.
Differences only appear in v0 of the Explorer API.

Close #152

### DIFF
--- a/ergo-lib/src/chain/ergo_box.rs
+++ b/ergo-lib/src/chain/ergo_box.rs
@@ -61,7 +61,7 @@ use std::convert::TryInto;
 #[cfg_attr(feature = "json", serde(try_from = "json::ergo_box::ErgoBoxFromJson"))]
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub struct ErgoBox {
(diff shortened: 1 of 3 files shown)


greenhat (approved):
Looking great! I appreciate your efforts!

PR #399, opened 2021-09-16
сontext headers

### LINKED ISSUE #372: `Contex.headers` and `Header` type
- `Context.headers` method in `scontext` module;
- `SType::SHeader` type;
- `sheader` module with Header properties (see `scontext` as example);

SabaunT:
Close #372

### DIFF
--- a/ergo-lib/src/wallet/signing.rs
+++ b/ergo-lib/src/wallet/signing.rs
@@ -131,7 +131,7 @@ pub fn make_context(
         inputs: inputs_ir,
         pre_header: state_ctx.pre_header.clone(),
         extension: ContextExtension::empty(),
-        headers: Vec::new()
+        headers: Vec::new(),
     })
 }
 
--- a/ergo-lib/src/wallet/signing.rs
+++ b/ergo-lib/src/wallet/signing.rs
@@ -131,6 +131,7 @@ pub fn make_context(
         inputs: inputs_ir,
         pre_header: state_ctx.pre_header.clone(),
         extension: ContextExtension::empty(),
+        headers: Vec::new()
     })
 }
 
--- a/ergotree-interpreter/src/eval/context.rs
+++ b/ergotree-interpreter/src/eval/context.rs
@@ -5,7 +5,7 @@ use std::rc::Rc;
 use crate::sigma_protocol::prover::ContextExtension;
 use ergotree_ir::ir_ergo_box::IrBoxId;
 use ergotree_ir::ir_ergo_box::IrErgoBoxArena;
-use ergotree_ir::mir::header::{PreHeader, Header};
+use ergotree_ir::mir::header::{Header, PreHeader};
 
 /// Interpreter's context (blockchain state)
 #[derive(Debug)]
@@ -65,7 +65,16 @@ mod arbitrary {
                 vec(any::<Header>(), 0..5),
             )
                 .prop_map(
-                    |(height, self_box, outputs, inputs, data_inputs, pre_header, extension, headers)| {
+                    |(
+                        height,
+                        self_box,
+                        outputs,
+                        inputs,
+                        data_inputs,
+                        pre_header,
+                        extension,
+                        headers,
+                    )| {
                         let self_box_id = self_box.id();
                         let outputs_ids = outputs.iter().map(|b| b.id()).collect();
                         let inputs_ids = inputs.iter().map(|b| b.id()).collect();
@@ -91,7 +100,7 @@ mod arbitrary {
                             inputs: inputs_ids,
                             pre_header,
                             extension,
-                            headers
+                            headers,
                         }
                     },
                 )
--- a/ergotree-interpreter/src/eval/context.rs
+++ b/ergotree-interpreter/src/eval/context.rs
@@ -5,7 +5,7 @@ use std::rc::Rc;
 use crate::sigma_protocol::prover::ContextExtension;
 use ergotree_ir::ir_ergo_box::IrBoxId;
 use ergotree_ir::ir_ergo_box::IrErgoBoxArena;
-use ergotree_ir::mir::header::PreHeader;
+use ergotree_ir::mir::header::{PreHeader, Header};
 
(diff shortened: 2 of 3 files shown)


ergo-lib/src/wallet/signing.rs
```diff
@@ -131,6 +131,7 @@ pub fn make_context(
         inputs: inputs_ir,
         pre_header: state_ctx.pre_header.clone(),
         extension: ContextExtension::empty(),
+        headers: Vec::new()
```
SabaunT, line 134: @greenhat Seems that need to add headers field to https://github.com/ergoplatform/sigma-rust/blob/6aa986d04b8955bd6c284ecf09954abb2c03f2a4/ergo-lib/src/chain/ergo_state_context.rs#L6, right?

greenhat: Yes, that's right.

greenhat:
Please, reference the issue in the description. `Close #372` seems appropriate in this case. This way you don't need to link them manually and the issue will be automatically closed when this PR is merged.
Also, including the issue number in the branch name would make it easy to identify the issue when you have only the branch name (git ops, etc.).

ergotree-interpreter/src/eval/context.rs
```diff
@@ -26,6 +26,8 @@ pub struct Context {
     pub pre_header: PreHeader,
     /// prover-defined key-value pairs, that may be used inside a script
     pub extension: ContextExtension,
+    /// Fixed number of last block headers in descending order (first header is the newest one)
+    pub headers: Vec<Header>,
```
greenhat, line 30: Since the number of headers is fixed `[Header; 10]` fits better here.

ergotree-ir/src/mir/header.rs
```diff
@@ -1,3 +1,5 @@
+use num_bigint::BigInt;
```
greenhat, line 1: `mir` (Middle Intermediate Representation) module is intended for IR nodes. New `chain` module for types specific to the blockchain seems like a good fit.

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
     pub votes: Vec<u8>,
 }
 
+/// Represents data of the block header available in Sigma propositions.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct Header {
+    /// Block version, to be increased on every soft and hardfork.
+    pub version: u8,
+    /// Bytes representation of ModifierId of this Header
+    pub id: Vec<u8>,
```
greenhat, line 31: Looks like a good fit for a new type. `BlockId(Digest32)` should be a good fit. See `TxId` for example. The same goes to `parent_id`.

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
...
+    /// Bytes representation of ModifierId of the parent block
+    pub parent_id: Vec<u8>,
+    /// Hash of ADProofs for transactions in a block
+    pub ad_proofs_root: Vec<u8>,
```
greenhat, line 35: `Digest32` should be enough for now, we'll add a specific domain type later when(if) we need to deal with AD proofs here. Check out ergo https://github.com/ergoplatform/ergo/blob/e68ce6180b13bffb024cf9e26c7c16a7be70a22c/src/main/scala/org/ergoplatform/modifiers/history/Header.scala#L57-L87 for more precise types.

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
...
+    /// AvlTree of a state after block application
+    pub state_root: AvlTree,
+    /// Root hash (for a Merkle tree) of transactions in a block.
+    pub transaction_root: Vec<u8>,
```
greenhat, line 39: `Digest32`

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
...
+    /// Timestamp of a block in ms from UNIX epoch
+    pub timestamp: u64,
+    /// Current difficulty in a compressed view.
+    pub n_bits: u64,
+    /// Block height
+    pub height: u32,
+    /// Root hash of extension section
+    pub extension_root: Vec<u8>,
```
greenhat, line 47: `Digest32`

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
...
+    /// One-time public key. Prevents revealing of miners secret.
+    pub pow_onetime_pk: Box<dlog_group::EcPoint>,
+    /// nonce
+    pub nonce: Vec<u8>,
+    /// Distance between pseudo-random number, corresponding to nonce `nonce` and a secret,
+    /// corresponding to `miner_pk`. The lower `pow_distance` is, the harder it was to find this solution.
+    pub pow_distance: BigInt,
+    /// Miner votes for changing system parameters.
+    /// 3 bytes in accordance to Scala implementation, but will use `Vec` until further improvements
+    pub votes: Vec<u8>,
```
greenhat, line 59: `[u8; 3]`?

ergotree-ir/src/mir/header.rs
```diff
@@ -20,6 +22,47 @@ pub struct PreHeader {
...
+}
+
+/// Temporary, until not implemented https://github.com/ergoplatform/sigma-rust/issues/368.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct AvlTree;
```
greenhat, line 64: Definitely deserves it's own module `avltree` in root.

greenhat (changes requested):
Congrats on the first contribution! Please check the comments.

greenhat:
A note regarding the types. We want to check as much as possible at the compile time rather than in runtime. Thus, we always want to use the most precise types possible.

PR #404, opened 2021-09-17
Convert `STypeCompanion` to enum

### LINKED ISSUE #339: Convert `STypeCompanion` to enum
### Motivation
- To have exhaust compiler check in `ergotree_interpreter::eval::smethod_eval_fn`;

https://github.com/ergoplatform/sigma-rust/blob/6698880353fbec1038d0947ca2488746bcb77f34/ergotree-interpreter/src/eval.rs#L176-L203

- To have exhaust compiler check instead of `ergotree_ir::types::stype_companion::STypeCompanion::type_by_id()`

https://github.com/ergoplatform/sigma-rust/blob/6698880353fbec1038d0947ca2488746bcb77f34/ergotree-ir/src/types/stype_companion.rs#L37-L56

kettlebell:
Close #339

### DIFF
--- a/ergotree-ir/src/types/stype_companion.rs
+++ b/ergotree-ir/src/types/stype_companion.rs
@@ -1,3 +1,4 @@
+use std::convert::TryFrom;
 use std::fmt::Debug;
 
 use crate::serialization::types::TypeCode;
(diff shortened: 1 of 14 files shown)


ergotree-ir/src/types/stype_companion.rs
```diff
@@ -22,66 +23,103 @@ pub(crate) struct STypeCompanionHead {
...
-    pub fn type_name(&'static self) -> &'static str {
-        self.head.type_name
+    pub fn type_name(&self) -> &'static str {
+        self.s_type_companion_head().type_name
+    }
+}
+
+impl TryFrom<TypeCode> for STypeCompanion {
+    type Error = SigmaParsingError;
+    fn try_from(value: TypeCode) -> Result<Self, Self::Error> {
```
kettlebell, line 102: @greenhat this code manually covers all the enum variants. We can get more compiler assurance by using the `strum` crate (https://crates.io/crates/strum), which has the ability to iterate over all enum variants. What do you think?

greenhat: `strum` is neat and I agree to pull it in. I'm wondering if we could ditch `STypeCompanionHead` completely. What do you think of adding `STypeCompanion::type_code()` to return type code and `STypeCompanion::type_name()` to return type name?

greenhat (changes requested):
Looking great! Please, check my suggestion.

ergotree-ir/src/types/stype_companion.rs
kettlebell: Sure, I'll make the changes.

greenhat (approved):
Looking great! Thank you!

Issue #402, opened 2021-09-17
[200 SigmaUSD] `Downcast` IR node for numeric types


greenhat:
Create a new struct in a separate module(file) in `mir` module and add it to `Expr` enum. 
Similar to the existing `Upcast` IR node.

Reference Scala implementation:
IR node - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c863a9b1a82589e47b15f76f3affdb30a475e740/sigmastate/src/main/scala/sigmastate/trees.scala#L568-L587
For evaluation implementation details see `SNumeric.downcast` https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c863a9b1a82589e47b15f76f3affdb30a475e740/sigmastate/src/main/scala/sigmastate/types.scala#L775-L781 implementations for numeric types (`SInt`, etc.)

ApexTheory:
I have started on this.

greenhat:
Done in #408

greenhat:
Bounty is sent. Thank you!

PR #408, opened 2021-09-21
Downcast IR node implementation

ApexTheory:
PR for https://github.com/ergoplatform/sigma-rust/issues/402

There were a few casts not implemented on the Scala side, so I left them out here as well.  For example:
BigInt -> BigInt/Long/Int/Short/Byte
Byte -> Short

Also. in the `eval` code downcast functions, I was wondering what your thoughts are on using match guards to check for overflows, rather than try_from. This way, it would match on the default error case and avoid the repeated error matching.

### DIFF
--- a/ergotree-interpreter/src/eval.rs
+++ b/ergotree-interpreter/src/eval.rs
@@ -50,6 +50,7 @@ pub(crate) mod create_provedlog;
 pub(crate) mod decode_point;
 mod deserialize_context;
 mod deserialize_register;
+pub(crate) mod downcast;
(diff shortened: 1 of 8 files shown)


greenhat (approved):
Looking great! I appreciate your efforts on test coverage!
Regarding the match with guards, I find the code fine as is. I assume code would be more compact using guards but I'm afraid rustc exhaustivity checks might not work in that case.

ApexTheory:
Converted to PR and ready for merge.

greenhat (approved):

PR #409, opened 2021-09-21
Impl `Box.bytesWithoutRef` method's IR node

### LINKED ISSUE #400: [150 SigmaUSD] `Box.bytesWithoutRef` method's IR node (`ExtractBytesWithNoRef`)
Create a new struct in a separate module(file) in `mir` module and add it to `Expr` enum. 
Similar to the existing `ExtractBytes` IR node but without the transaction id and the index in the output.

### Serialization
- Implement `Arbitrary` trait and add serialization roundtrip PBT test for this node;

### Evaluation
- Implement `Evaluable` trait in `ergotree-interpreter` crate;
- Add new method in `IrErgoBox::bytesWithoutRef`
- Add evaluation tests;

Check out the existing IR node implementation (`ExtractBytes`) for details.

Check out the detailed implementation instructions for `Append` op in  #301

Reference Scala version:
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/c863a9b1a82589e47b15f76f3affdb30a475e740/sigmastate/src/main/scala/sigmastate/utxo/transformers.scala#L437-L456

ross-weir:
Closes #400

### DIFF
--- a/ergo-lib/src/chain/ergo_box.rs
+++ b/ergo-lib/src/chain/ergo_box.rs
@@ -20,6 +20,8 @@ use ergotree_ir::serialization::SigmaSerializeResult;
 use ergotree_ir::util::AsVecI8;
 pub use register::*;
 
+use sigma_util::DIGEST32_SIZE;
+
 #[cfg(feature = "json")]
 use super::json;
 use super::token::TokenAmount;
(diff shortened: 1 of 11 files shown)


ergo-lib/src/chain/ergo_box.rs
```diff
@@ -214,6 +216,14 @@ impl IrErgoBox for ErgoBox {
     fn bytes(&self) -> Result<Vec<i8>, SigmaSerializationError> {
         Ok(self.sigma_serialize_bytes()?.as_vec_i8())
     }
+
+    fn bytes_without_ref(&self) -> Result<Vec<i8>, SigmaSerializationError> {
```
ross-weir, line 220: The implementation at the time of the commit linked in the issue used `ErgoBoxCandidate::bytes` with a `TODO`: 

```
// TODO v5.0: re-implement extracting directly from `ErgoBox.bytes` array
```

Figured I'd give this solution a shot, not sure if it's right.

greenhat (changes requested):
Looking great! Please see my comments.

greenhat (approved):
Great! Thank you!

PR #433, opened 2021-10-02
Evaluation for Header properties

### LINKED ISSUE #421: [300 SigmaUSD] Evaluation for `Header` properties
from #418
For each Header property implement evaluation in `eval::sheader` module.

See `sbox`, `scoll` modules for examples.

SabaunT:
Close #421

### DIFF
--- a/ergotree-interpreter/src/eval/sheader.rs
+++ b/ergotree-interpreter/src/eval/sheader.rs
@@ -27,10 +27,10 @@ mod tests {
     use std::convert::TryInto;
     use std::rc::Rc;
 
+    use ergotree_ir::chain::block_id::BlockId;
     use ergotree_ir::mir::{coll_by_index::ByIndex, expr::Expr, property_call::PropertyCall};
     use ergotree_ir::types::{scontext, sheader, smethod::SMethod};
     use ergotree_ir::util::AsVecU8;
-    use ergotree_ir::chain::block_id::BlockId;
     use sigma_test_util::force_any_val;
 
     use crate::eval::{context::Context, tests::eval_out};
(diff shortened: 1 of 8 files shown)


ergotree-ir/src/types/sheader.rs
```diff
@@ -36,146 +66,97 @@ lazy_static! {
...
 }
 
-pub const VOTES_PROPERTY_METHOD_ID: MethodId = MethodId(15);
 lazy_static! {
-    pub static ref VOTES_PROPERTY: SMethod =
-        SMethod::new(Header, VOTES_PROPERTY_METHOD_DESC.clone(),);
-    static ref VOTES_PROPERTY_METHOD_DESC: SMethodDesc =
-        property("votes", SColl(SByte.into()), VOTES_PROPERTY_METHOD_ID);
+    /// Property used only for tests!!!
+    pub static ref UNKNOWN_PROPERTY: SMethod = SMethod::new(Header, property("unknown", SType::SByte, MethodId(100)),);
```
greenhat, line 159: I suggest moving it to eval::sheader::tests module.

ergotree-interpreter/src/eval/sheader.rs
```diff
@@ -0,0 +1,350 @@
...
+            )))
+        )
+    }
+
+    #[test]
+    fn test_eval_failed_unknown_property() {
+        let expr = create_get_header_property_expr(sheader::UNKNOWN_PROPERTY.clone());
+        assert_eq!(
+            try_eval_out_wo_ctx::<i8>(&expr),
+            Err(EvalError::NotFound(format!(
```
greenhat, line 344: I'd love to catch the exact error in such cases myself. However, this makes tests too fragile. Think of changing the error message wording braking the test. 
I just check if its `is_err()` until a better (non-fragile) solution is found.

greenhat (changes requested):
Looking great! Check my minor comments.

ergotree-ir/src/types/sheader.rs
SabaunT: I should think about some workaround here, because `SMethod` has one private field and private for other crates constructors.
Or can I make `pub` what it's needed to move this to the stated module?

greenhat: Oh, please make `SMethod::new` constructor fully public then. I see no reason why I made it that way.

PR #467, opened 2021-10-27
Audit unwraps and panics in `ergo-lib`

### LINKED ISSUE #265: [500 SigmaUSD] Ban (via clippy) todo!, panic!, unimplemented! in `ergo-lib` crate
Each case should be reviewed and explicitly allowed. 

Add:
```
#![deny(clippy::unwrap_used)]
#![deny(clippy::expect_used)]
#![deny(clippy::todo)]
#![deny(clippy::panic)]
```
at the root `lib.rs`.

For reference and examples check out how it was done in the following creates before:
- ~sigma-ser~ #328;
- ~ergotree-ir~ #328;
  - extend ‘sigma_serialize‘ error besides io error (use type alias for Result);
- ~ergotree-interpreter~ #344;

I'll be happy to assist since some cases will be hard to figure out without the knowledge of the inner workings of Ergo chain type.

**EDIT: Not all unwraps should be eliminated. The goal is to eliminate those that are "purely runtime" by their nature and "uncheckable" at compile-time and/or in tests. It sounds a bit vague, so please check https://github.com/ergoplatform/sigma-rust/pull/461 for some examples. In cases where it's not clear feel free to ping me.**

kettlebell:
Closes #265.

In summary I've only removed unwraps relating to `TokenAmount::checked_[add|sub]`. For the other non-obvious unwraps (to me) I've added a comment explaining why it's safe.

### DIFF
--- a/ergo-lib/src/wallet/box_selector/simple.rs
+++ b/ergo-lib/src/wallet/box_selector/simple.rs
@@ -45,7 +45,8 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
         let mut selected_boxes_value: u64 = 0;
         let target_balance: u64 = target_balance.into();
         // sum all target tokens into hash map (think repeating token ids)
-        let mut target_tokens_left: HashMap<TokenId, TokenAmount> = sum_tokens(Some(target_tokens));
+        let mut target_tokens_left: HashMap<TokenId, TokenAmount> =
+            sum_tokens(Some(target_tokens)).map_err(BoxSelectorError::TokenAmountError)?;
         let mut has_value_change = false;
         let mut has_token_change = false;
         let mut sorted_inputs = inputs;
@@ -66,7 +67,7 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
         });
         // reverse, so they'll be sorted by descending order (boxes with target tokens will be first)
         sorted_inputs.reverse();
-        sorted_inputs.into_iter().for_each(|b| {
+        for b in sorted_inputs {
             let value_change_amt: u64 = if target_balance > selected_boxes_value {
                 0
             } else {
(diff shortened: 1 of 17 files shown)


ergo-lib/src/wallet/box_selector/simple.rs
```diff
@@ -45,7 +45,8 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
         let mut selected_boxes_value: u64 = 0;
         let target_balance: u64 = target_balance.into();
         // sum all target tokens into hash map (think repeating token ids)
-        let mut target_tokens_left: HashMap<TokenId, TokenAmount> = sum_tokens(Some(target_tokens));
+        let mut target_tokens_left: HashMap<TokenId, TokenAmount> =
+            sum_tokens(Some(target_tokens)).map_err(BoxSelectorError::TokenAmountError)?;
```
greenhat, line 49: It can be simplified to:
```suggestion
            sum_tokens(Some(target_tokens))?;
```
if `From` is implemented like in https://github.com/ergoplatform/sigma-rust/blob/86e012029ffab9590ef0898ff98f11bd30b8b004/ergotree-interpreter/src/eval.rs#L137 (with `use thiserror::Error;
` imported).

greenhat (changes requested):
Looking great! Please check my comments on how to ditch `map_err` and allow unwrap for expression/statement only.

ergo-lib/src/wallet/box_selector/simple.rs
kettlebell: Nice thanks!

ergo-lib/src/wallet/tx_builder.rs
```diff
@@ -199,8 +200,10 @@ impl<S: ErgoBoxAssets + ErgoBoxId + Clone> TxBuilder<S> {
             ));
         }
         // check that inputs have enough tokens
-        let input_tokens = sum_tokens_from_boxes(self.box_selection.boxes.as_slice());
-        let output_tokens = sum_tokens_from_boxes(output_candidates.as_slice());
+        let input_tokens = sum_tokens_from_boxes(self.box_selection.boxes.as_slice())
+            .map_err(TxBuilderError::TokenAmountError)?;
+        let output_tokens = sum_tokens_from_boxes(output_candidates.as_slice())
+            .map_err(TxBuilderError::TokenAmountError)?;
```
greenhat, line 206: The same `map_err` -> `?` is applicable here.

greenhat (changes requested):
Great! See missing `map_err` -> `?` conversion in tx_builder.

ergo-lib/src/wallet/tx_builder.rs
kettlebell: Oops, sorry I missed that.

greenhat (approved):

PR #469, opened 2021-10-31
changing hints bag and adding add_hint function

sahandzou:
For accessing hints inside HintsBag I need to make hints public.

For adding a new hint to the HintsBag I wrote the add_hint function like the Scala version. if it is accepted I will add another function for adding Vector of hints to the HintsBag.


greenhat:
> For accessing hints inside HintsBag I need to make hints public.
> 
> For adding a new hint to the HintsBag I wrote the add_hint function like the Scala version. if it is accepted I will add another function for adding Vector of hints to the HintsBag.

Looking good!

Issue #367, opened 2021-08-10
[500 SigmaUSD] Distributed signatures support


greenhat:
References:
[EIP-11](https://github.com/ergoplatform/eips/pull/8)
[Improved distributed signatures](https://www.ergoforum.org/t/improved-distributed-signatures/366)

New code should be in `prover::multi_sig` module.

#### Generate commitments
`fn generate_commitments(tx_context: TransactionContext, state_context: &ErgoStateContext) -> Result<TransactionHintsBag, _> ` function should be an equivalent of `wallet/generateCommitments` node API call, see implementation at https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoProvingInterpreter.scala#L182-L222
`TransactionHintsBag` type should be an equivalent of https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/TransactionHintsBag.scala#L5

#### Extract hints from signed(invalid) transaction
`fn extract_hints(tx_context: TransactionContext, state_context: &ErgoStateContext, real_secrets_to_extract: Vec<SigmaBoolean>, simulated_secrets_to_extract: Vec<SigmaBoolean>) -> Result<TransactionHintsBag, _>` should be an equivalent of `script/extractHints` node API call, see implementation at https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoProvingInterpreter.scala#L224-L263

#### Hints support in prover and tx signing
- Check how hints are used in prover and implement the missing parts compared to the Scala reference implementation at https://github.com/ScorexFoundation/sigmastate-interpreter/blob/455d4a3b5a82eec9301be89019d8ea675386474a/sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala#L68-L103
- add `tx_hints: TransactionHintsBag` parameter to `Wallet::sign_*` methods and pass it down to prover (as `HintsBag`);

#### Wasm API. 
- Expose `TransactionHintsBag` type via wrapper;
- Add `tx_hints: TransactionHintsBag` parameters to `Wallet::sign_*` methods;
[...]

PR #470, opened 2021-11-01
Multi sig

sahandzou:
This is my draft for ```fn generate_commitments(tx_context: TransactionContext, state_context: &ErgoStateContext)``` it still need some refactor and commenting.

I have some questions:

1- For generating commitments with respect to the private keys that prover have the ```fn generate_commitments_for(
    sigmatree: SigmaBoolean,
    generate_for: &Vec<SigmaBoolean>,
) -> HintsBag``` function getting a ```Vec<SigmaBoolean>``` as an input. my question is where should I implement this function according to my understanding of the Scala version this function should be proverInterpreter I want some tips about it to where I should locate this function?

2- I need some tips about guidelines for tests in sigma-rust; for example, what should be covered in this code?

3- Anything I did not consider in my code that should be corrected.


greenhat:
> This my draft for `fn generate_commitments(tx_context: TransactionContext, state_context: &ErgoStateContext)` it still need some refactor and commenting.

Great! 

> I have some questions:
> 
> 1- For generating commitments with respect to the private keys that prover have the `fn generate_commitments_for( sigmatree: SigmaBoolean, generate_for: &Vec<SigmaBoolean>, ) -> HintsBag` function get a `Vec<SigmaBoolean>` as an input. my question is where should I implement this function according to my understanding of the Scala version this function should be proverInterpreter I want some tips about it to where I should locate this function?

`prover::multi_sig` should be a good fit.

> 2- I need some tips about guidelines for tests in sigma-rust such for example what should be covered in this code?

Check out prover spec tests in Scala (see Testing in #367 for details). We should strive to cover the same cases.

> 3- Anything I did not consider in my code that should be corrected.

Nothing caught my eye so far besides formatting and some names for local variables.

sahandzou:
Hey @greenhat 
I think it's done I also test the wasm binding and test it with the main net and it was ok.

For implementation, I should be made some struct public and it causes to "missing doc" linter error so I suppress that. Could you please take a look at that and tell me what I should do?

greenhat:
> Hey @greenhat I think it's done I also test the wasm binding and test it with the main net and it was ok.

Great! Please resolve conflicts.
> 
> For implementation, I should be made some struct public and it causes to "missing doc" linter error so I suppress that. Could you please take a look at that and tell me what I should do?

Please, don't suppress the linter warning for missing docs. Every external/public API should be documented. Consider `pub(crate)` for items that are for "internal" use inside the crate. Also, feel free to copy and amend doc comments from Scala implementation where it makes sense.

sahandzou:
> > Hey @greenhat I think it's done I also test the wasm binding and test it with the main net and it was ok.
> 
> Great! Please resolve conflicts.
> 
> > For implementation, I should be made some struct public and it causes to "missing doc" linter error so I suppress that. Could you please take a look at that and tell me what I should do?
> 
> Please, don't suppress the linter warning for missing docs. Every external/public API should be documented. Consider `pub(crate)` for items that are for "internal" use inside the crate. Also, feel free to copy and amend doc comments from Scala implementation where it makes sense.

I fixed the issues. ready for review!

bindings/ergo-lib-wasm/src/transaction/reduced.rs
```diff
@@ -68,6 +81,99 @@ impl ReducedTransaction {
...
+            .first()
+            .clone()
+            .reduction_result
+            .sigma_prop
+            .sigma_serialize_bytes()
+            .map_err(to_js)
+    }
+
+    /// Generate commitment for first input with input address
+    pub fn generate_commitment_for_first_input(
```
greenhat, line 98: Why do we need it? I don't see it is used anywhere in the code? The same question also goes to `get_first_input_serialized`

sahandzou: Now should be fixed.

bindings/ergo-lib-wasm/src/transaction/reduced.rs
```diff
@@ -64,10 +77,86 @@ impl ReducedTransaction {
...
+            }
+            serde_json::to_string_pretty(&commitments)
+                .map_err(|e| JsValue::from_str(&format!("{}", e)))
+        } else {
+            Err(JsValue::from_str("index is out of bound"))
+        }
+    }
+
+    /// bag for multi sig
+    pub fn bag_for_multi_sig(
```
greenhat, line 117: Why do we need this?

bindings/ergo-lib-wasm/src/transaction/reduced.rs
```diff
@@ -64,10 +77,86 @@ impl ReducedTransaction {
...
     }
 
-    /// Returns the unsigned transation
+    /// Returns the unsigned transaction
     pub fn unsigned_tx(&self) -> UnsignedTransaction {
         self.0.unsigned_tx.clone().into()
     }
+
+    /// Generate commitment for index input with input address
+    pub fn generate_commitment_for_input(
```
greenhat, line 86: Why do we need this? I don't see it's used anywhere.

bindings/ergo-lib-wasm/src/wallet.rs
```diff
@@ -40,6 +45,7 @@ impl Wallet {
         tx: &UnsignedTransaction,
         boxes_to_spend: &ErgoBoxes,
         data_boxes: &ErgoBoxes,
+        // tx_hints: &TransactionHintsBag,
```
greenhat, line 48:
```suggestion
```

bindings/ergo-lib-wasm/src/wallet.rs
```diff
@@ -62,6 +68,42 @@ impl Wallet {
             .map(Transaction::from)
     }
 
+    /// Sign a transaction:
```
greenhat, line 71: This doc comment is the same as for `sign_transaction` and does not convey what differences does it have.

bindings/ergo-lib-wasm/src/wallet.rs
```diff
@@ -62,6 +68,42 @@ impl Wallet {
...
+        &self,
+        _state_context: &ErgoStateContext,
+        tx: &UnsignedTransaction,
+        boxes_to_spend: &ErgoBoxes,
+        data_boxes: &ErgoBoxes,
+        hints: &HintsBag,
+    ) -> Result<Transaction, JsValue> {
+        let boxes_to_spend = TxIoVec::from_vec(boxes_to_spend.clone().into()).map_err(to_js)?;
+        let mut tx_hints = TransactionHintsBag::empty();
+        tx_hints.add_hints_for_input(0, hints);
```
greenhat, line 86: Why does it handle hints for the first input only?

bindings/ergo-lib-wasm/src/transaction/reduced.rs
greenhat: Thanks!

ergo-lib/src/wallet/signing.rs
```diff
@@ -180,13 +182,51 @@ pub fn sign_transaction(
...
+    })?;
+    Ok(Transaction::new(
+        signed_inputs,
+        tx.data_inputs,
+        tx.output_candidates,
+    )?)
+}
+
+/// Signs a multi sig transaction (generating proofs for inputs)
+pub fn sign_transaction_multi(
```
greenhat, line 205: Is there any reason to have tx signing with hints as a separate function? I'd rather add `tx_hints: Option<TransactionHintsBag>` parameter to `sign_transaction`. The same goes for `Wallet::sign_transaction_multi` (for Wasm layer we have to keep both `sign_transaction` and `sign_transaction_multi` since there is no `Option` in JS).

bindings/ergo-lib-wasm/src/wallet.rs
sahandzou: I Merged the multi_sig_signer with the signer and solved this issue.

ergo-lib/src/wallet/signing.rs
sahandzou: I merged the two functions even in the Wasm layer. the `TransactionHintsBag` exposed `empty` method in the Wasm layer that should be passed to `sign_transaction`; I think this resolves "multiple signers" in the Wasm layer. Is that ok or should I keep both signers?

bindings/ergo-lib-wasm/src/wallet.rs
sahandzou: By merging signers this issue is resolved.

bindings/ergo-lib-wasm/src/transaction/reduced.rs
sahandzou: This function returns commitments(Secret & Public) that are generated for specific input boxes in the Wasm layer.
I used it for generating commitments for inputs. Should I change it to a function that generates commitments for a whole transaction?

bindings/ergo-lib-wasm/tests/test_transaction.js
```diff
@@ -184,7 +185,8 @@ it('use signed tx outputs as inputs in a new tx', async () => {
   const sks = new SecretKeys();
   sks.add(sk);
   const wallet = Wallet.from_secrets(sks);
-  const signed_tx = wallet.sign_transaction(ctx, tx, unspent_boxes, tx_data_inputs);
+  const tx_hints_bag = TransactionHintsBag.empty();
+  const signed_tx = wallet.sign_transaction(ctx, tx, unspent_boxes, tx_data_inputs, tx_hints_bag);
```
sahandzou, line 189: Updated Js test of New sign_transaction in Wasm layer.

ergo-lib/src/wallet/signing.rs
greenhat: We as library authors should strive to keep our API easy to understand (think of newcomers). For the majority of our user's cases, they would be signing a regular tx and forcing them to learn what does `TransactionHintsBag` means and how they have to pass an `emtpy` one is an unnecessary burden. Thus, it's better to encode this optional value with types, i.e. `Option<TransactionHintsBag>`. This way, a simple glance through the docs and passing familiar `None` will be the only "price" for the majority of our users. Unfortunately, JS does not have the `Option` type, thus we need to have two `sign_transaction` and `sign_transaction_multi` in Wasm API.

bindings/ergo-lib-wasm/src/transaction/reduced.rs
greenhat: This method is in Wasm layer and I don't see it's used anywhere. Why does it take a secret key as a parameter?

greenhat: And yes, we're missing commitment generation for non-reduced transactions.

greenhat: As a side note, Wasm layer is intended to be a thin and simple "conversion" layer for JS bindings. Don't forget about Rust(+ C, Java, Swift, bindings) users of this library. The core features should be implemented in ergo-lib crate so that other bindings could use it as well.

ergo-lib/src/wallet/signing.rs
sahandzou: Absolutely right! I am trying in new commits to consider this point and I think now it's done.

bindings/ergo-lib-wasm/src/transaction/reduced.rs
sahandzou: That was a good point! I should be considered this point, I am tried in new commits considering this point and refactor the codes. I think now it's done and ready to review.

sahandzou:
Hi @greenhat 

I refactored the code and I think is ready to review.

bindings/ergo-lib-wasm/src/transaction.rs
```diff
@@ -20,10 +20,154 @@ use std::convert::TryInto;
...
+}
+
+/// Generate commitments for transaction
+#[wasm_bindgen]
+pub fn generate_commitments(
+    _state_context: &ErgoStateContext,
+    tx: &UnsignedTransaction,
+    boxes_to_spend: &ErgoBoxes,
+    data_boxes: &ErgoBoxes,
+    public_keys: Propositions,
```
greenhat, line 144: What do you think of moving `generate_commitments` from `Transaction` and `ReducedTransaction` to `Wallet`? This will eliminate the need of `public_keys` parameter since wallet already has them.

greenhat:
> Hi @greenhat
> 
> I refactored the code and I think is ready to review.

Hey! Good job! Thank you for your patience! Please see my comment. I'm focusing on step 1 (commitment generation) now. When we're done with it let's jump to tx signing and the to the hint extraction. This way we will progress step-by-step through the whole PR.

bindings/ergo-lib-wasm/src/transaction.rs
sahandzou: Yes, good point! I moved them in the last commit.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,844 @@
...
+
+#[allow(clippy::unwrap_used)]
+/// Generate commitments for transaction
+pub fn generate_commitments(
+    tx_context: TransactionContext,
+    state_context: &ErgoStateContext,
+    public_keys: &[SigmaBoolean],
+) -> Result<TransactionHintsBag, TxSigningError> {
+    let mut hints_bag = TransactionHintsBag::empty();
+    for (i, input) in tx_context.get_boxes_to_spend().enumerate() {
```
greenhat, line 154: `tx_context.get_boxes_to_spent()` are not the same as tx inputs. Although there must be a box in `boxes_to_spend` for every tx inputs the boxes might be in a different order.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,844 @@
...
+        ut,
+        real_propositions,
+        simulated_propositions,
+        NodePosition::crypto_tree_prefix(),
+        &mut bag,
+    );
+    bag
+}
+
+#[allow(clippy::unwrap_used)]
```
greenhat, line 146: We should handle all the errors here including `tree.proposition()`.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,844 @@
...
+/// Generate commitments for transaction
```
greenhat, line 147: I believe this doc comment can be better. You can adopt one from the equivalent method in Scala - https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoProvingInterpreter.scala#L182-L194

greenhat: Also applies to `generate_commitments_for`

greenhat:
Hey! Good job on commitment generation changes! Besides the above comments, it looks good!

ergo-lib/src/wallet/multi_sig.rs
sahandzou: I write better doc comments for both of them in the new commit.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: I fixed the unwrapped use in tree proposition in the new commit.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Yes, I didn't know that was fixed. i handled it in the new commit.

sahandzou:
> Hey! Good job on commitment generation changes! Besides the above comments, it looks good!

Thanks, I resolve them in the new commit ready to go for the next step!

ergo-lib/src/wallet/signing.rs
```diff
@@ -170,6 +171,7 @@ pub fn sign_transaction(
     prover: &dyn Prover,
     tx_context: TransactionContext,
     state_context: &ErgoStateContext,
+    tx_hints: &TransactionHintsBag,
```
greenhat, line 174: Let's make this `tx_hints: Option<&TransactionHintsBah>` as well. As it is done in `wallet`.

greenhat, line 174: We're missing a code path for signing a reduced transaction with hints. Let's add the same optional parameter as in `sign_transaction` above and in `Wallet::sign_transaction` and the second method in Wasm `Wallet::sign_reduced_transaction_multi`.

sahandzou:
> > > Hey! Good job on commitment generation changes! Besides the above comments, it looks good!
> > 
> > 
> > Thanks, I resolve them in the new commit ready to go for the next step!
> 
> Well done! Transaction signing looks good besides my notes above.

Thanks, I resolve them ready for the next step!

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+        let sigma_tree = reduction_result.sigma_prop;
+        hints_bag.add_hints_for_input(i, generate_commitments_for(sigma_tree, public_keys));
+    }
+    Ok(hints_bag)
+}
+
+/// Extracting hints from a transaction and outputs it's corresponding TransactionHintsBag
+pub fn extract_hints(
+    signed_tx: Transaction,
+    tx_context: TransactionContext,
```
greenhat, line 182: Since `TransactionContext` holds an unsigned tx which has no place here let's make `TransactionContext<T>` polymorphic in `spending_tx` type to allow both signed and unsigned transactions. Thus, eliminating the need for `signed_tx` parameter here. 
Hint: To make it polymorphic you're going to need a new `ErgoTransaction` trait to make `TransactionContext::new` work.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+            ),
+        );
+    }
+
+    Ok(hints_bag)
+}
+
+#[allow(clippy::unwrap_used)]
+/// Extracting hints from reduced transaction
+pub fn extract_hints_from_reduced_transaction(
```
greenhat, line 228: Why do we need hint extraction for reduced tx? It's an unsigned tx while hint extraction requires a signed tx.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+            .map_err(|e| TxSigningError::ProverError(e, i))?;
+
+        let sigma_tree = reduction_result.sigma_prop;
+        hints_bag.add_hints_for_input(i, generate_commitments_for(sigma_tree, public_keys));
+    }
+    Ok(hints_bag)
+}
+
+/// Extracting hints from a transaction and outputs it's corresponding TransactionHintsBag
+pub fn extract_hints(
```
greenhat, line 180: We need to expose `extract_hints` as function in Wasm API. And since we don't have `TransactionContext` in Wasm we need to pass boxes to spend and data boxes directly in there.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+        );
+        hints_bag.add_hints_for_input(idx, hints);
+    }
+    hints_bag
+}
+
+// The unwrap use on option causes panic in case of dhtuple signature, that is not implemented
+#[allow(clippy::unwrap_used)]
+/// Traversing node of sigma tree
+pub fn traverse_node(
```
greenhat, line 260: Let's move it inside `bag_for_multi_sig` function.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+            let comm = compute_commitment(&proposition, &challenge, &second_message);
+            ret = Some(FirstDlogProverMessage::from(comm));
+        }
+        UncheckedLeaf::UncheckedDhTuple(_) => {}
+    }
+    ret
+}
+
+#[allow(clippy::unwrap_used)]
+/// Outputs Hints bag for a signed(invalid) transaction
```
greenhat, line 127: I believe doc comment could be better. We can copy from https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/.metals/readonly/dependencies/sigma-state_2.12-4.0.3-sources.jar/sigmastate/interpreter/ProverUtils.scala#L64-L76

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+            let second_message = pdlog.second_message;
+            let comm = compute_commitment(&proposition, &challenge, &second_message);
+            ret = Some(FirstDlogProverMessage::from(comm));
+        }
+        UncheckedLeaf::UncheckedDhTuple(_) => {}
+    }
+    ret
+}
+
+#[allow(clippy::unwrap_used)]
```
greenhat, line 126: We should refrain from unwraps here and handle all the errors.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,862 @@
...
+            simulated_propositions.as_slice(),
+            proof.as_slice(),
+        );
+        hints_bag.add_hints_for_input(idx, hints);
+    }
+    hints_bag
+}
+
+// The unwrap use on option causes panic in case of dhtuple signature, that is not implemented
+#[allow(clippy::unwrap_used)]
```
greenhat, line 258: Please, handle the errors without unwrapping here. In general, unwraps besides tests are allowed for exceptional cases only.

greenhat:
Good job on transaction signing! Please, check out my comments for hints extraction.

sahandzou:
> Good job on transaction signing! Please, check out my comments for hints extraction.

Thanks! I think now the problems should be fixed!

ergo-lib/src/wallet/signing.rs
```diff
@@ -49,11 +49,23 @@ pub struct TransactionContext {
     data_boxes: Option<TxIoVec<ErgoBox>>,
 }
 
-impl TransactionContext {
+/// Ergo Transaction that made TransactionContext from Transaction and UnsignedTransaction
+pub trait ErgoTransaction<T> {
```
greenhat, line 53: It does not make much sense for the transaction to construct its context from itself and additional pieces of information (boxes) that it's not responsible for.

I was talking about something like this:
```rust

pub struct TransactionContext<T: ErgoTransaction> {
    /// Transaction (signed or unsigned) that spending `boxes_to_spend` (e.g has them in it's inputs) and has `data_boxes` in it's data inputs
    pub spending_tx: T,
    /// Boxes corresponding to tx inputs
    boxes_to_spend: TxIoVec<ErgoBox>,
    /// Boxes corresponding to tx data inputs
    data_boxes: Option<TxIoVec<ErgoBox>>,
}

impl<T: ErgoTransaction> TransactionContext<T> {

    fn new(
        spending_tx: T,
        boxes_to_spend: TxIoVec<ErgoBox>,
        data_boxes: Option<TxIoVec<ErgoBox>>,
        ) -> Result<Self, TxSigningError> {
        /// impl against generic tx (ErgoTransaction trait)
    }

}

pub trait ErgoTransaction {
    pub fn input_ids(&self) -> TxIoVec<BoxId>;
    pub fn data_inputs(&self) -> Option<TxIoVec<DataInput>>;
}

impl ErgoTransaction for UnsignedTransaction {
    ...
}

impl ErgoTransaction for Transaction {
    ...
}
```

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,850 @@
...
+    proof: &[u8],
+) -> Result<HintsBag, &'static str> {
+    let ut = parse_sig_compute_challenges(&sigma_tree, proof.to_owned());
+    match ut {
+        Ok(tree) => {
+            let mut bag: HintsBag = HintsBag::empty();
+
+            // The unwrap use on option causes panic in case of dhtuple signature, that is not implemented
+            // Traversing node of sigma tree
+            fn traverse_node(
```
greenhat, line 143: No need to increase the indentation level for `traverse_node` and intertwine it inside the "main" code. Let's move it to the beginning of the `bag_for_multi_sig`.

ergo-lib/src/wallet/signing.rs
sahandzou: That's right, I fix it in the new commit.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Right! i made the change in the new commit.

sahandzou:
Hey @greenhat
any update on the PR state?

ergo-lib/src/wallet/signing.rs
```diff
@@ -97,63 +81,35 @@ impl ErgoTransaction<UnsignedTransaction> for TransactionContext {
...
-    ) -> Result<Self, TxSigningError> {
-        for (i, unsigned_input) in spending_tx.inputs.iter().enumerate() {
-            if !boxes_to_spend
-                .iter()
-                .any(|b| unsigned_input.box_id == b.box_id())
-            {
-                return Err(TxSigningError::InputBoxNotFound(i));
-            }
-        }
+/// ErgoTransaction trait implemented for `Transaction` & `UnsignedTransaction`
```
greenhat, line 84:
```suggestion
/// Exposes common properties for signed and unsigned transactions
```

ergo-lib/src/chain/transaction.rs
```diff
@@ -55,7 +55,8 @@ use std::iter::FromIterator;
 )]
 #[derive(PartialEq, Debug, Clone)]
 pub struct Transaction {
-    tx_id: TxId,
+    /// transaction id
+    pub tx_id: TxId,
```
greenhat, line 59: The reason `tx_id` is private is that it's a computed property (hash of the serialized tx) and should not be set externally. Why do you want it to be public?

greenhat:
Transaction signing looks good! See minor notes.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,906 @@
...
+
+        let sigma_tree = reduction_result.sigma_prop;
+        hints_bag.add_hints_for_input(i, generate_commitments_for(sigma_tree, public_keys));
+    }
+    Ok(hints_bag)
+}
+
+/// Extracting hints from a transaction and outputs it's corresponding TransactionHintsBag
+pub fn extract_hints(
+    tx: &Transaction,
```
greenhat, line 262: Let's use the updated `TransactionContext<Transaction>` instead of tx, boxes and data boxes.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,906 @@
...
+            .get(i)
+            .ok_or_else(|| {
+                TxSigningError::ContextError(
+                    "self_index not found in spending transaction inputs".to_string(),
+                )
+            })?
+            .spending_proof
+            .extension
+            .clone();
+        let ctx = Rc::new(Context {
```
greenhat, line 326: Why not use `signing::make_context` here?

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,906 @@
...
+/// A method which is extracting partial proofs of secret knowledge for particular secrets with their
+/// respective public images given. Useful for distributed signature applications.
+/// See DistributedSigSpecification for examples of usage.
+pub fn bag_for_multi_sig(
+    sigma_tree: SigmaBoolean,
+    real_propositions: &[SigmaBoolean],
+    simulated_propositions: &[SigmaBoolean],
+    proof: &[u8],
+) -> Result<HintsBag, &'static str> {
+    let ut = parse_sig_compute_challenges(&sigma_tree, proof.to_owned());
```
greenhat, line 138: Let's change the return type error to `SigParsingError` and then after
```suggestion
    let ut = parse_sig_compute_challenges(&sigma_tree, proof.to_owned())?;
```
remove all the error handling in the end of the this function.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,906 @@
...
+        let reduction_result = reduce_to_crypto(&exp, &Env::empty(), ctx)
+            .map_err(ProverError::EvalError)
+            .map_err(|e| TxSigningError::ProverError(e, i))?;
+        let sigma_tree = reduction_result.sigma_prop;
+        let bag = bag_for_multi_sig(
+            sigma_tree,
+            real_secrets_to_extract.as_slice(),
+            simulated_secrets_to_extract.as_slice(),
+            proof.as_slice(),
+        );
```
greenhat, line 352: Let's properly propagate the error up with
```suggestion
        )?;
```
and remove `match` below which silence the error.

greenhat:
Hint extraction looks good besides my notes. 

Hey! I appreciate your patience and willingness to make changes! Keep up the good work!

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Yes, good point I rewrite that in the new commits!

ergo-lib/src/wallet/multi_sig.rs
sahandzou: We talk about the problem that occurs I manually create `Context` for now

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Yes, changed that!

ergo-lib/src/wallet/multi_sig.rs
sahandzou: fixed!

sahandzou:
> Hint extraction looks good besides my notes.
> 
> Hey! I appreciate your patience and willingness to make changes! Keep up the good work!

Thank You! the comments are considered in the new commits; ready for review!

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -349,7 +354,10 @@ pub fn extract_hints(
             real_secrets_to_extract.as_slice(),
             simulated_secrets_to_extract.as_slice(),
             proof.as_slice(),
-        );
+        )
+        .map_err(|_e| {
+            TxSigningError::ProverError(Unexpected("Unexpected Sigma tree Value".to_string()), i)
+        });
```
greenhat, line 360:
```suggestion
        })?;
        hints_bag.add_hints_for_input(i, bag);
```
and you don't need the following `match` which silences the error case anyway.

ergo-lib/Cargo.toml
```diff
@@ -29,10 +29,12 @@ thiserror = "1"
 derive_more = "0.99"
 # bounded-vec = { git = "https://github.com/ergoplatform/bounded-vec", rev="f684d78" }
 bounded-vec = { version = "^0.5.0" }
-num-bigint = "0.4.0"
+num-bigint = { version = "0.4.0", features = ["serde"] }
 proptest-derive = {version = "0.3.0", optional = true }
 sha2 = { version = "0.9.8" }
 hmac = { version = "0.11.0" }
+k256 = { version = "0.10.0", features = ["arithmetic", "ecdsa"] }
+hex = "0.4.3"
```
greenhat, line 37: We're already using `base16` library for hex encoding. Do you have a reason to prefer `hex` over `base16`?

ergo-lib/src/chain/json/hints.rs
```diff
@@ -0,0 +1,313 @@
...
+use k256::elliptic_curve::ops::Reduce;
+use k256::{Scalar, U256};
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+
+/// OwnCommitmentJson serialization
+#[derive(Serialize, Deserialize)]
+pub struct OwnCommitmentJson {
+    /// secret
+    pub secret: String,
```
greenhat, line 25: 1. Why do you need these structs and their fields to be public?
2. Doc comments do not add any value. If they are internal types doc comments are not necessary.

greenhat: Wait, these `*CommitmentJson` structs are not used anywhere. Delete?

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,914 @@
...
+            proof.as_slice(),
+        )
+        .unwrap();
+        assert!(!bag.hints.is_empty(), "{}", "{}");
+    }
+
+    #[test]
+    fn generating_commitment_two_signer() {
+        let secret1 = DlogProverInput::random();
+        let secret2 = DlogProverInput::random();
```
greenhat, line 573: `secret2` should be `DhTupleProverInput` according to 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/test/scala/sigmastate/utxo/ProverSpecification.scala#L22

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,914 @@
...
+        }
+        hint = bag.hints[1].clone();
+        if let Hint::CommitmentHint(CommitmentHint::OwnCommitment(comm)) = hint {
+            assert_eq!(
+                comm.position,
+                NodePosition {
+                    positions: vec![0, 0]
+                }
+            );
+        }
```
greenhat, line 670: DH case is missing according to https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/test/scala/sigmastate/utxo/ProverSpecification.scala#L49-L57

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -0,0 +1,914 @@
...
+                    ctx,
+                    proof_byte2,
+                    message.as_slice()
+                )
+                .unwrap()
+                .result,
+            "{}",
+            "{}"
+        );
+    }
```
greenhat, line 913: Although many scenarios from the Scala spec are using `AtLeast` conjecture which we don't have implemented yet there is a scenario that we're missing here - 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/test/scala/sigmastate/utxo/DistributedSigSpecification.scala#L391-L431 or-and-or with both dlog and dht.

greenhat:
Hey! Hint extraction is looking good! I've gone through tests and noted some missing cases. 
Besides that, we need a smoke test for the JS/Wasm API for a simple two-party distributed signing procedure (similar to the one we have in the `multi-sig` module). See https://github.com/sahandzou/sigma-rust/blob/5e6bb25e9bc25b9e4c0d8cde3d75792bf719eefe/bindings/ergo-lib-wasm/tests/test_transaction.js#L43-L76
tx signing test as an example. The goal is to test JS/Wasm API for any missing parts needed in the whole distributed signing process on the JS side.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Good Point! i changed it.

ergo-lib/Cargo.toml
sahandzou: After new changes, there is no need for `hex` anymore so I remove it

ergo-lib/src/chain/json/hints.rs
sahandzou: Yes I delete that.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: i changed it in new commits

ergo-lib/src/wallet/multi_sig.rs
sahandzou: My bad i changed them in new commit.

ergo-lib/src/wallet/multi_sig.rs
sahandzou: I added new test in new commits

sahandzou:
Hey! I added the missing tests and also the Wasm tests for two-party distributed signing and doing some bug fix now ready for review!

ergotree-interpreter/src/sigma_protocol/prover.rs
```diff
@@ -1010,11 +1008,29 @@ fn step9_real_dh_tuple<P: Prover + ?Sized>(
...
+            {
+                Some(proof) => {
+                    let unchecked_tree = proof.clone().unchecked_tree;
+                    if let UncheckedTree::UncheckedLeaf(UncheckedLeaf::UncheckedDhTuple(
+                        unchecked_dht,
+                    )) = unchecked_tree
+                    {
+                        unchecked_dht.second_message
+                    } else {
+                        let z =
```
greenhat, line 1024: According to Scala reference implementation in this case we should return an error. See https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala#L535
`asInstanceOf` will just throw if casting is impossible.

ergotree-interpreter/src/sigma_protocol/unchecked_tree.rs
```diff
@@ -116,6 +130,12 @@ impl UncheckedSchnorr {
     pub fn with_challenge(self, challenge: Challenge) -> Self {
         UncheckedSchnorr { challenge, ..self }
     }
+    /// compute commitment
+    pub fn compute_commitment(&self) -> FirstDlogProverMessage {
```
greenhat, line 134: 1. Why do we need to compute commitments here and for DH tuple below and not use `self.commitmentOpt`?
2. Why `verifier.compute_commitments()` call is missing in `bag_for_multisig` after https://github.com/sahandzou/sigma-rust/blob/7b1aeea527bc05a24d0c868fe24b5b7f6f14aa54/ergo-lib/src/wallet/multi_sig.rs#L121 ?

greenhat:
> Hey! I added the missing tests and also the Wasm tests for two-party distributed signing and doing some bug fix now ready for review!

Great! I'm sorry for the late replay. Good job! Please see my questions above.

ergotree-interpreter/src/sigma_protocol/prover.rs
sahandzou: Yes Right, i think in the new commit the problem is solved.

sahandzou:
> > Hey! I added the missing tests and also the Wasm tests for two-party distributed signing and doing some bug fix now ready for review!
> 
> Great! I'm sorry for the late reply. Good job! Please see my questions above.

Thank You! in the new commits the problems told above are fixed.

ergo-lib/src/wallet/multi_sig.rs
```diff
@@ -147,26 +147,30 @@ pub fn bag_for_multi_sig(
...
+                                SecretProven::RealSecretProof(RealSecretProof {
+                                    image: leaf.proposition(),
+                                    challenge: leaf.challenge(),
+                                    unchecked_tree: UncheckedTree::UncheckedLeaf(leaf),
+                                    position,
+                                }),
+                            );
+                            bag.add_hint(real_commitment);
+                            bag.add_hint(real_secret_proof);
+                        }
```
greenhat, line 168: In case of `leaf.commitment_opt()` is None we should throw an error. And in the case below as well. I suggest moving commitment extraction above `if real_found {...} else {...}` expression to avoid the repeat.

ergotree-interpreter/src/sigma_protocol/prover.rs
```diff
@@ -1024,9 +1024,7 @@ fn step9_real_dh_tuple<P: Prover + ?Sized>(
                     {
                         unchecked_dht.second_message
                     } else {
-                        let z =
-                            dlog_group::random_scalar_in_group_range(crypto_utils::secure_rng());
-                        SecondDhTupleProverMessage { z }
+                        return Err(ProverError::SecretNotFound);
```
greenhat, line 1027:
```suggestion
                return Err(ProverError::Unexpected(format!(
                    "Expected unchecked DH tuple in proof.unchecked_tree, got {:?}",
                    unchecked_tree
                )));
```

ergotree-interpreter/src/sigma_protocol.rs
```diff
@@ -35,7 +35,7 @@ use derive_more::From;
 use derive_more::TryInto;
 
 /** The message sent by a prover to its associated verifier as part of a sigma protocol interaction. */
-pub(crate) trait ProverMessage {
+pub trait ProverMessage {
```
greenhat, line 38: Are you sure `ProverMessage` should be public? I don't see where it's used outside the crate.

ergotree-interpreter/src/sigma_protocol/prover.rs
```diff
@@ -892,24 +896,74 @@ fn step9_real_schnorr<P: Prover + ?Sized>(
...
-            Ok(Some(
-                UncheckedSchnorr {
-                    proposition: us.proposition.clone(),
-                    commitment_opt: None,
-                    challenge,
-                    second_message: z,
-                }
-                .into(),
-            ))
+            let oc = hints_bag
```
greenhat, line 899: I suggest more idiomatic Rust approach to handling optional values. Consider the whole block inside `if let Some(challenge) ...` can be rewritten as:
```rust

        let z: SecondDlogProverMessage = if let Some(priv_key) = prover
            .secrets()
            .iter()
            .flat_map(|s| match s {
                PrivateInput::DlogProverInput(dl) => vec![dl],
                _ => vec![],
            })
            .find(|prover_input| prover_input.public_image() == us.proposition)
        {
            hints_bag
                .own_commitments()
                .into_iter()
                .find(|comm| comm.position == us.position)
                .map(|oc| {
                    dlog_protocol::interactive_prover::second_message(
                        priv_key,
                        oc.secret_randomness,
                        &challenge,
                    )
                })
                .unwrap_or({
                    dlog_protocol::interactive_prover::second_message(
                        priv_key,
                        us.randomness_opt.ok_or_else(|| {
                            ProverError::Unexpected(format!("empty randomness in {:?}", us))
                        })?,
                        &challenge,
                    )
                })
        } else {
            match hints_bag
                .real_proofs()
                .into_iter()
                .find(|comm| comm.position == us.position)
            {
                Some(tree) => {
                    let unchecked_tree = tree.unchecked_tree;
                    if let UncheckedTree::UncheckedLeaf(UncheckedLeaf::UncheckedSchnorr(
                        unchecked_schnorr,
                    )) = unchecked_tree
                    {
                        unchecked_schnorr.second_message
                    } else {
                        return Err(ProverError::SecretNotFound);
                    }
                }
                None => {
                    let bs = dlog_group::random_scalar_in_group_range(crypto_utils::secure_rng());
                    SecondDlogProverMessage { z: bs }
                }
            }
        };
        Ok(Some(
            UncheckedSchnorr {
                proposition: us.proposition,
                commitment_opt: None,
                challenge,
                second_message: z,
            }
            .into(),
        ))
```

greenhat (changes requested):
Looking good! Great job! I think we're almost done here. Please check my final notes and let's merge it afterward!

ergo-lib/src/wallet/multi_sig.rs
sahandzou: Yes i moving commitment extraction to the above of `if` block

ergotree-interpreter/src/sigma_protocol.rs
sahandzou: OW, Yes I missed that

ergotree-interpreter/src/sigma_protocol/prover.rs
sahandzou: I rewrite the block same as dht case

sahandzou:
> Looking good! Great job! I think we're almost done here. Please check my final notes and let's merge it afterward!

Thank You! I fixed them.

greenhat (approved):
Great job! I appreciate your patience! Thank you!

PR #471, opened 2021-11-02
BoxSelection::boxes cannot be empty

### LINKED ISSUE #463: [50 SigmaUSD] `BoxSelection::boxes` cannot be empty
### Motivation
In the error-less box selection `BoxSelection::boxes` cannot be empty. At least one box is required. 

### Solution
Use `BoundedVec` to ensure non-empty `BoxSelection::boxes`.
`pub type SelectedBoxes<T> = BoundedVec<T, 1, { u16::MAX as usize }>;` should suffice since tx inputs are limited to `u16::MAX`.

Venoox:
Closes #463 

I still need to solve this in wasm bindings, but before I continue, would this approach be good?

### DIFF
--- a/ergo-lib/src/wallet/box_selector.rs
+++ b/ergo-lib/src/wallet/box_selector.rs
@@ -65,18 +65,12 @@ pub enum BoxSelectorError {
 
     /// Boxes out of bounds
     #[error("Boxes is out of bounds")]
-    OutOfBounds(BoundedVecOutOfBounds),
+    OutOfBounds(#[from] BoundedVecOutOfBounds),
 }
 
 impl From<BoxValueError> for BoxSelectorError {
     fn from(e: BoxValueError) -> Self {
         BoxSelectorError::BoxValueError(e)
-    }
-}
-
-impl From<BoundedVecOutOfBounds> for BoxSelectorError {
-    fn from(e: BoundedVecOutOfBounds) -> Self {
-        BoxSelectorError::OutOfBounds(e)
     }
 }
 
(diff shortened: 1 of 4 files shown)


ergo-lib/src/wallet/tx_builder.rs
```diff
@@ -332,7 +326,7 @@ mod tests {
     #[test]
     fn test_empty_inputs() {
```
Venoox, line 327: I guess this test should be restructured since tx builder won't be returning EmptyInputBoxSelection error anymore but it will fail right away when trying to convert empty vec into bounded vec

ergo-lib/src/wallet/box_selector.rs
```diff
@@ -58,6 +62,10 @@ pub enum BoxSelectorError {
     /// Token amount err
     #[error("TokenAmountError: {0:?}")]
     TokenAmountError(#[from] TokenAmountError),
+
+    /// Boxes out of bounds
+    #[error("Boxes is out of bounds")]
+    OutOfBounds(BoundedVecOutOfBounds),
```
greenhat, line 68:
```suggestion
    OutOfBounds(#[from] BoundedVecOutOfBounds),
```

should remove the need for manual `From` impl below.

ergo-lib/src/wallet/tx_builder.rs
greenhat: Actually, this test does not make sense in tx builder anymore. Neither in BoxSelection. It's a BoxSelector test now, when there are not enough assets to select but I think we already have that covered.

greenhat:
> Closes #463
> 
> I still need to solve this in wasm bindings, but before I continue, would this approach be good?

Yes, it looks great!

greenhat (approved):
Looking great! Thank you!

PR #472, opened 2021-11-07
`flatMap` allowed ops check in lambda body

### LINKED ISSUE #222: Check for allowed ops in Coll.flatmap lambda
See the check in sigmastate - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/2cb966eac2c763e3aa7f344ca1c77384fb1fc056/sigmastate/src/main/scala/sigmastate/types.scala#L1423-L1496

Should be checked on Coll.flatmap evaluation.

ross-weir:
closes #222


greenhat (approved):
Looking great! Thank you!

PR #476, opened 2021-11-18
[ergotree-interpreter] Convert upcast/downcast tests to proptest

### LINKED ISSUE #474: [50 SigmaUSD] Fix `Downcast` eval tests and convert them to proptest along with `Upcast`
`to_byte` test failed on CI - https://github.com/ergoplatform/sigma-rust/runs/4202100586?check_suite_focus=true . I presume it was due to `v_byte` being zero. 

- Fix the `to_byte` test for this case and check that other tests do not exhibit this behavior;
- Convert `Downcast` and `Upcast` test to proptest;

SethDusek:
Hi, this PR converts the tests for both Upcast/Downcast to proptest. I've also added a new overflow test which confirms that overflow fails when downcasting a value that's out of bounds.

Closes #474

### DIFF
--- a/ergotree-interpreter/src/eval/downcast.rs
+++ b/ergotree-interpreter/src/eval/downcast.rs
@@ -124,7 +124,6 @@ impl Evaluable for Downcast {
 #[cfg(test)]
 mod tests {
     use ergotree_ir::mir::constant::Constant;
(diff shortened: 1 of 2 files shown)


greenhat (approved):
Looking great! I appreciate added overflow tests.

Issue #248, opened 2021-04-29
[300 SigmaUSD] Sigma protocol implementation for AtLeast


greenhat:
Implemented in #377:
- ~~evaluation(exept `Cthreshold::reduce`);~~
- ~~prover(exept Gf2_192*);~~
- ~~verifier;~~

Todo:
- #473
- ~~implement `Cthreshold::reduce` [Scala reference impl](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/262a14d17ae3090fa97d1de1effd0b338562e4db/sigmastate/src/main/scala/sigmastate/trees.scala#L467-L518);~~ #482
- ~~test verifier with proofs from mainnet/sigmastate [Scala test vector](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/262a14d17ae3090fa97d1de1effd0b338562e4db/sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala#L119-L135);~~ #535 
- test prover with verifier by generating arbitrary `AtLeast` IR node proving it and then passing through the verifier (see `verifier::tests` for examples);

ross-weir:
I'm starting on this one 👍

greenhat:
> I'm starting on this one 👍

Great! Please be aware that #473 is needed for testing prover/verifier.

greenhat:
@ross-weir Hey! #473 is implemented and merged. Everything is in place to finish this task.

ross-weir:
> @ross-weir Hey! #473 is implemented and merged. Everything is in place to finish this task.

Thanks, restarting this one

greenhat:
The bounty is sent. Thank you!

PR #482, opened 2021-11-20
`Cthreshold.reduce` impl

ross-weir:
Completes task from #248 

`implement Cthreshold::reduce Scala reference impl;`


greenhat (approved):
Looking good! 

Although I'm not entirely convinced by comments allowing unwraps :) I can only find unrealistic cases like `AtLeast(2, [PK, false])` where unwraps should fail. I believe that for now, keeping this code in sync with Scala version is more important than handling such cases.

PR #484, opened 2021-11-21
add binding for ErgoBox in wasm

minotaur-ergo:


bindings/ergo-lib-wasm/package.json
```diff
@@ -27,5 +27,7 @@
         "webpack-cli": "^3.1.1",
         "webpack-dev-server": "^3.1.0"
     },
-    "dependencies": {}
+    "dependencies": {
+        "wasm-pack": "^0.10.1"
```
greenhat, line 31: Why it's needed here? I mean it seems like a wrapper for a wasm-pack binary from what I see in https://github.com/rustwasm/wasm-pack/tree/master/npm

greenhat (changes requested):
Looking good! See my question on adding a dependency.

bindings/ergo-lib-wasm/package.json
minotaur-ergo: It is required for all build processes locally. as you can see in your package.json for example build-browser command used wasm-pack to build project.
my mistake is to install it as dependency. but it is a devDependency.

minotaur-ergo: I think Its required to to installed as devDependency to build project without GA. but removed it from project for now.

greenhat (approved):
Great! Thank you!

PR #492, opened 2021-11-24
Wallet from mnemonic impl

ross-weir:
Adds `Wallet.from_mnemonic`

~~Added a dependency to assist with `pbkdf2` as I didn't see anything in the codebase already for that, chose [`ring`](https://github.com/briansmith/ring) because it looks pretty active.~~

Added RustCrypto dependency for `pbkdf2`

`expected` seed bytes in tests where taken from [ `AddressGenerationDemo`](https://github.com/ergoplatform/ergo/blob/master/ergo-wallet/src/test/java/org/ergoplatform/wallet/AddressGenerationDemo.java#L16) to confirm implementation

### DIFF
--- a/bindings/ergo-lib-wasm/src/wallet.rs
+++ b/bindings/ergo-lib-wasm/src/wallet.rs
@@ -1,4 +1,5 @@
 //! Wallet-like features
+use derive_more::{From, Into};
 use ergo_lib::chain::transaction::TxIoVec;
 use wasm_bindgen::prelude::*;
 
@@ -13,6 +14,7 @@ use crate::{
 
 /// A collection of secret keys. This simplified signing by matching the secret keys to the correct inputs automatically.
 #[wasm_bindgen]
+#[derive(From, Into)]
 pub struct Wallet(ergo_lib::wallet::Wallet);
 
 #[wasm_bindgen]
@@ -75,11 +77,5 @@ impl Wallet {
             .sign_reduced_transaction(reduced_tx.clone().into())
             .map_err(to_js)
             .map(Transaction::from)
-    }
-}
-
-impl From<ergo_lib::wallet::Wallet> for Wallet {
-    fn from(t: ergo_lib::wallet::Wallet) -> Self {
-        Wallet(t)
     }
 }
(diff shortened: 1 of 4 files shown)


greenhat:
Quick note regarding `ring`. Although I also heard good things about it we're knees deep in RustCrypto ecosystem, so I'd rather use https://github.com/RustCrypto/password-hashes

ergo-lib/src/wallet.rs
```diff
@@ -41,6 +43,16 @@ impl From<TxSigningError> for WalletError {
 }
 
 impl Wallet {
+    /// Create wallet instance loading secret key from mnemonic
+    pub fn from_mnemonic(mnemonic_phrase: &str, mnemonic_pass: &str) -> Wallet {
```
ross-weir, line 47: I think this is likely wrong. I have a feeling that we might need to implement `ExtendedPrivateKey` but would love to get clarification before going down that rabbit hole @greenhat 😄 

My other feeling is extended private keys might only be needed when we start generating addresses so maybe this is OK for just signing transactions etc

greenhat: No, this is fine for the start. It looks similar to https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/nodeView/wallet/ErgoWalletService.scala#L283-L299 

And yes, we're going to need ext private keys when start generating(deriving) addresses.

bindings/ergo-lib-wasm/src/wallet.rs
```diff
@@ -76,3 +77,9 @@ impl Wallet {
             .map(Transaction::from)
     }
 }
+
+impl From<ergo_lib::wallet::Wallet> for Wallet {
```
greenhat, line 81: `From` can be derived. See https://github.com/ergoplatform/sigma-rust/blob/ed132f81eaa612b39da3c5c180f244b458522916/bindings/ergo-lib-wasm/src/transaction.rs#L29-L30

greenhat (changes requested):
Looking great! See a minor note.

greenhat (approved):
Great! Thank you!

PR #493, opened 2021-11-26
Add peer management

### LINKED ISSUE #479: [200 SigmaUSD] Peer management
Create the following types in ergo-p2p crate:
- `PeerSpec`
Ergo node implementation - https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/scorex/core/network/PeerSpec.scala#L11-L27

- `PeerInfo`
Ergo node implementation - https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/scorex/core/network/peer/PeerInfo.scala#L8-L17

Implement serialization for the types above by implementing `ScorexSerializable` trait from sigma-ser (after #477 is implemented).

Create `PeerDatabase` trait and `InMemoryPeerDatabase` implementation with the following methods:
- `get_peer_by_addr`;
- `add_or_update_peer`;
- `remove_peer`;
- `known_peers`;

See Ergo node implementation - https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/scorex/core/network/peer/InMemoryPeerDatabase.scala#L14 for details. Ignore blacklisting and penalizing peers methods since we don't need them now.

ross-weir:
closes #479

### DIFF
--- a/ergo-p2p/src/lib.rs
+++ b/ergo-p2p/src/lib.rs
@@ -17,6 +17,7 @@
 #![deny(clippy::unimplemented)]
 #![deny(clippy::panic)]
 
+pub mod peer_addr;
 pub mod peer_feature;
 pub mod peer_spec;
 pub mod protocol_version;
--- a/ergo-p2p/src/lib.rs
+++ b/ergo-p2p/src/lib.rs
@@ -17,9 +17,16 @@
 #![deny(clippy::unimplemented)]
 #![deny(clippy::panic)]
 
-pub mod peer_addr;
-pub mod peer_database;
-pub mod peer_feature;
-pub mod peer_info;
-pub mod peer_spec;
-pub mod protocol_version;
+mod peer_addr;
+mod peer_database;
+mod peer_feature;
+mod peer_info;
+mod peer_spec;
+mod protocol_version;
+
+pub use peer_addr::PeerAddr;
+pub use peer_database::{in_memory::InMemoryPeerDatabase, PeerDatabase, PeerDatabaseError};
+pub use peer_feature::{LocalAddressPeerFeature, PeerFeature, PeerFeatureId};
+pub use peer_info::PeerInfo;
+pub use peer_spec::PeerSpec;
+pub use protocol_version::ProtocolVersion;
--- a/ergo-p2p/src/lib.rs
+++ b/ergo-p2p/src/lib.rs
@@ -18,6 +18,8 @@
 #![deny(clippy::panic)]
 
 pub mod peer_addr;
+pub mod peer_database;
 pub mod peer_feature;
+pub mod peer_info;
 pub mod peer_spec;
 pub mod protocol_version;
--- a/ergo-p2p/src/lib.rs
+++ b/ergo-p2p/src/lib.rs
@@ -16,3 +16,7 @@
 #![deny(clippy::todo)]
 #![deny(clippy::unimplemented)]
 #![deny(clippy::panic)]
+
+pub mod peer_feature;
+pub mod peer_spec;
+pub mod protocol_version;
--- a/ergo-p2p/src/peer_addr.rs
+++ b/ergo-p2p/src/peer_addr.rs
@@ -74,6 +74,7 @@ pub mod arbitrary {
 #[allow(clippy::panic)]
 #[allow(clippy::unwrap_used)]
 #[cfg(test)]
+#[cfg(feature = "arbitrary")]
 mod tests {
     use super::*;
     use proptest::prelude::*;
--- a/ergo-p2p/src/peer_addr.rs
+++ b/ergo-p2p/src/peer_addr.rs
@@ -71,6 +71,8 @@ pub mod arbitrary {
     }
 }
 
+#[allow(clippy::panic)]
+#[allow(clippy::unwrap_used)]
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/ergo-p2p/src/peer_addr.rs
+++ b/ergo-p2p/src/peer_addr.rs
@@ -4,12 +4,12 @@ use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
 };
 
-use derive_more::{From, Into};
+use derive_more::{Display, From, Into};
 use sigma_ser::{ScorexSerializable, ScorexSerializationError};
 
 /// Peer address
-#[derive(PartialEq, Eq, Debug, Copy, Clone, From, Into, Hash)]
-pub struct PeerAddr(pub SocketAddr);
+#[derive(PartialEq, Eq, Debug, Copy, Clone, From, Into, Hash, Display)]
+pub struct PeerAddr(SocketAddr);
 
 impl PeerAddr {
     /// Size in bytes of the ip address associated with this peer address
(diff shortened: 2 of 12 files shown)


ross-weir:
@greenhat Still working on this but if you had some spare time I'd appreciate a light look over to make sure I'm not doing anything horribly wrong 😄 

Particularly unsure about `PeerSpec` serialization, it feels quite verbose at the moment, not sure if there's better ways

greenhat:
> @greenhat Still working on this but if you had some spare time I'd appreciate a light look over to make sure I'm not doing anything horribly 

You're doing great! Keep it up.

ergo-p2p/src/peer_spec.rs
```diff
@@ -0,0 +1,117 @@
...
+                io::Error::new(io::ErrorKind::InvalidInput, "failed to parse ip size")
+            })?;
+
+            w.put_u8(addr_size)?;
+            addr.scorex_serialize(w)?;
+
+            Ok(())
+        })?;
+
+        // Can't use Vec<ScorexSerializable> becuase we need the size as u8
```
ross-weir, line 76: I'm actually not sure if this is the case, as it seems the underlying `put_u` functions cast everything to 64 sized ints

ross-weir: Had a bit more of a look in the Scala codebase and it looks like that's part of the VLQ encoding

greenhat: Although you're right that `put_u32` for 0..256 values will be equivalent to `put_u8` I'd rather use `put_u8` to serialize elements manually. In the protocol, it's a one-byte so let's explicitly encode it in the type system. Actually, let's go one step further and make "illegal states unrepresentable" :). What do you think of making `PeerSpec::features` be of type `Option<BoundedVec<1, 255, PeerFeature>>`?

ross-weir: Yep sounds good, I was a bit loose with the serialization types because of the upcasting so I'll go back through and make sure they're updated to the correct sizes 👍

ergo-p2p/src/peer_addr.rs
```diff
@@ -0,0 +1,64 @@
...
+#[allow(clippy::panic)]
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use sigma_ser::scorex_serialize_roundtrip;
+
+    #[test]
+    fn ser_roundtrip() {
+        let obj: PeerAddr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 8080).into();
+        assert_eq![scorex_serialize_roundtrip(&obj), obj]
```
greenhat, line 62: Although it's not strictly needed for this test I think we should start `Arbitrary` implementations for all relevant types early and use proptests. Let's place it in the separate `arbitrary` module like in ergotree-ir crate so that we could export them behind the feature later.

ergo-p2p/src/peer_database.rs
```diff
@@ -0,0 +1,43 @@
+//! Peer database types
+
+use std::collections::HashMap;
+
+use crate::{peer_addr::PeerAddr, peer_info::PeerInfo};
+
+type PeerMap = HashMap<PeerAddr, PeerInfo>;
+
+trait PeerDatabase {
```
greenhat, line 9:
```suggestion
pub trait PeerDatabase {
```
?

ergo-p2p/src/peer_database.rs
```diff
@@ -0,0 +1,43 @@
...
+        self.peers.get(&addr)
+    }
+
+    fn add_or_update_peer(&mut self, peer_info: PeerInfo) {
+        if let Some(addr) = peer_info.peer_spec.addr() {
+            self.peers.insert(addr, peer_info);
+        }
+    }
+
+    fn remove_peer(&mut self, addr: PeerAddr) {
```
greenhat, line 36: We should probably return an error in case a peer is not found. 
Although `InMemoryPeerDatabase` is tiny and simple we should cover it at least with smoke tests to catch the regressions.

ergo-p2p/src/peer_addr.rs
```diff
@@ -0,0 +1,64 @@
...
+    convert::TryInto,
+    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
+};
+
+use derive_more::{From, Into};
+use sigma_ser::{ScorexSerializable, ScorexSerializationError};
+
+/// Peer address
+#[derive(PartialEq, Eq, Debug, Copy, Clone, From, Into, Hash)]
+pub struct PeerAddr(pub SocketAddr);
```
greenhat, line 12: I'd refrain from making the inner item public. It kinda goes against the idea of newtype and leaks implementation details. `.into()` should suffice for users outside of the module.

ergo-p2p/src/protocol_version.rs
```diff
@@ -0,0 +1,52 @@
+//! Ergo P2P network version
+use sigma_ser::{ScorexSerializable, ScorexSerializeResult};
+
+/// P2P network protocol version
+#[derive(PartialEq, Eq, Debug, Clone, Hash)]
+pub struct ProtocolVersion(pub u8, pub u8, pub u8);
```
greenhat, line 6: Named fields?

greenhat (changes requested):
Looking great! Please, check my comments.

ergo-p2p/src/lib.rs
```diff
@@ -16,3 +16,10 @@
 #![deny(clippy::todo)]
 #![deny(clippy::unimplemented)]
 #![deny(clippy::panic)]
+
+pub mod peer_addr;
```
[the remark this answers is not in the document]
ross-weir: Ohh that's really nice, I like it 👍

ergo-p2p/src/peer_addr.rs
```diff
@@ -0,0 +1,90 @@
...
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            (any::<[u8; 4]>(), any::<u16>())
+                .prop_map(|(octets, port)| {
+                    SocketAddr::new(Ipv4Addr::from(octets).into(), port).into()
+                })
+                .boxed()
+        }
+    }
+}
+
```
greenhat, line 73:
```suggestion

#[cfg(feature = "arbitrary")]
```
Should help. Or, alternatively maybe adding `proptest` to `dev-dependencies` will fix it as well.

ross-weir: Neither done the trick, here's what I think is happening:

When running tests in a workspace if one dependency enables a feature then it's enabled across all of them for all their tests:
- cargo test --package ergotree-ir fails
- cargo test passes because ergo-lib enables the arbitrary feature: ergotree-ir = { version = "^0.14.0", path = "../ergotree-ir", features = ["arbitrary"] }

Tests are passing now because they're behind the `arbitrary` feature, which I think makes sense because the tests are completely dependent on `Arbitrary`. I think the way to go is adding `--features=arbitrary` to the ci action?

greenhat: Thanks for digging into this! I tested and `ergotree-ir` crate behaves the same way. I'm not a fan of adding `--features=arbitrary`. I'd prefer tests run without any additional options. I found this workaround - https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481 What do you think?

ross-weir: I agree with not needing to enable additional options for tests, added the workaround 👍

ergo-p2p/src/peer_feature.rs
```diff
@@ -0,0 +1,146 @@
...
+use derive_more::{From, Into};
+
+use sigma_ser::vlq_encode::WriteSigmaVlqExt;
+use sigma_ser::{ScorexParsingError, ScorexSerializable, ScorexSerializeResult};
+
+use crate::peer_addr::PeerAddr;
+
+/// Peer feature identifier
+#[derive(PartialEq, Eq, Debug, Copy, Clone, From, Into)]
+pub struct PeerFeatureId(pub u8);
```
greenhat, line 13:
```suggestion
pub struct PeerFeatureId(u8);
```

greenhat (changes requested):
Looking great! Thank you! Just a tiny comment/suggestion.

greenhat (approved):
Great! Thank you!

Issue #47, opened 2020-06-17
[5000 SigmaUSD] C + Swift(iOS) bindings 


greenhat:
## Scope
Implement an API feature-wise on par with Wasm bindings (ergo-lib-wasm). Basically an API that a wallet would need to be able to send and receive coins and tokens including input box selection, output box creation and transaction building and signing.

## Architecture
The core part is the common/shared source code with JNI (ergo-lib-jni) that should be implemented in ergo-lib-c-core crate.
The generic C bindings should be implemented in ergo-lib-c which should be used in Swift wrapper that should be implemented in ergo-lib-ios.

### C
The main idea is to pass Rust types wrapped in opaque pointers and expose functions to manipulate them on C side.
The crucial part is to free resources on the same side(usually Rust) where they were allocated. For this we need to expose a separate function for every "resource-producing" function that frees a resource passed from C. In case of Address see address_from_testnet and address_delete.

Use the already implemented Address type as an example. 

### Swift
Wraps functions from ergo-lib-c in classes for resource management and exposes methods converting to/from Swift types where necessary.

## Types
Using Wasm bindings as a reference we need the following types(with methods) implemented. For implmentation details see Wasm types and their methods in ergo-lib-wasm. Some method names are adopted for C types. See implemented Address type

Wallet:
from_secrets
sign_transaction

SecretKey:
dlog_from_bytes
to_bytes

ErgoStateContext:
new(from PreHeader)

PreHeader:
from_block_header

BlockHeader:
from_json

BlockHeaders:
from_json

UnsignedTransaction:
id
inputs
data_inputs
output_candidates
to_json
from_json

ErgoBox:
new
box_id
creation_height
tokens
ergo_tree
value
register_value
to_json
from_json

ErgoBoxCandidate:
creation_height
tokens
ergo_tree
value
register_value

Token:
new
id
amount

TokenId:
from_box_id
from_str
to_str

TokenAmount:
from_int64_t
as_int64_t

BoxValue:
from_int64_t
as_int64_t

Transaction:
id
to_json
from_json
inputs
data_inputs
output_candidates
outputs

DataInput:
new
box_id

TxBuilder:
new
set_data_inputs
build
box_selection
data_inputs
output_candidates
current_height
fee_amount
change_address
min_change_value

ErgoBoxCandidateBuilder:
new
set_min_box_value_per_byte
min_box_value_per_byte
set_value
value
calc_box_size_bytes
calc_min_box_value
set_register_value
register_value
[...]

PR #494, opened 2021-12-01
Fix 'signed tx as input' test in `ergo-lib-wasm`

kettlebell:
Fix the test as discussed in https://github.com/ergoplatform/sigma-rust/issues/47#issuecomment-983380015


greenhat (approved):
Thank you for taking care of this! I appreciate it!

PR #498, opened 2021-12-07
Start extended secret key

ross-weir:
Start work on `ExtendedSecretKey` implementation


ergo-lib/src/wallet/ext_secret_key.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/// Private key (serialized Scalar) bytes
+pub type SecretKeyBytes = [u8; 32];
+/// Chain code bytes
+pub type ChainCode = [u8; 32];
+
+type HmacSha512 = Hmac<Sha512>;
+
+/// Result type for ExtendedSecretKey operations
+pub type Result<T> = std::result::Result<T, ExtSecretKeyError>;
```
greenhat, line 27: I'd refrain from this. It's a trade-off between speed of writing and readability and I find readability more important.

ergo-lib/src/wallet/ext_secret_key.rs
```diff
@@ -0,0 +1,267 @@
...
+        Ok(ExtPubKey::new(
+            // unwrap is safe as it is used on an Infallible result type
+            self.public_image_bytes()?.try_into().unwrap(),
+            self.chain_code,
+            self.derivation_path.clone(),
+        )?)
+    }
+
+    /// Derive a child extended secret key using the provided index
+    #[allow(clippy::unwrap_used)]
```
greenhat, line 102: I prefer keeping `allow` directive above the actual call. It saves you hunting for it inside the function and disincentivizes blanket `allow`.

ergo-lib/src/wallet/ext_secret_key.rs
```diff
@@ -0,0 +1,267 @@
...
+            chain_code.try_into().unwrap(),
+            DerivationPath::master_path(),
+        )
+        .unwrap();
+
+        for v in vectors {
+            ext_secret_key = ext_secret_key.derive(v.next_index).unwrap();
+            assert_eq!(ext_secret_key.secret_key_bytes(), v.expected_secret_key);
+        }
+    }
```
greenhat, line 266: Could you add test vectors from the node wallet implementation from https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKeySpec.scala#L18-L50
Just to be extra sure that they are equivalent.

ross-weir: Good call on this, it made me realize our extended key interfaces might be a bit off compared to Scala.

See 8b8e9ff , the main thing being what was previously `derive` should be `child`, there should be a `derive` method that takes a full path instead of an index.

If you agree with the above I'd be happy to fix-up `ext_pub_key` and maybe extract things out into a `ext_key` trait like in node: https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedKey.scala#L16

greenhat: > Good call on this, it made me realize our extended key interfaces might be a bit off compared to Scala.
> 
> See [here](https://github.com/ergoplatform/sigma-rust/commit/8b8e9ff85889d1771fbb024c18f8486b5cc1f618#diff-cc91791574e7a77dd357ac1364ee2c7bf71ace4a776d91b964cd7b774dc20344R102-R149), the main thing being what was previously `derive` should be `child`, there should be a `derive` method that takes a full path instead of an index.

Good catch on `child`! I don't see a `derive` with full path in node wallet but it's up to you.
 
> If you agree with the above I'd be happy to fix-up `ext_pub_key` and maybe extract things out into a `ext_key` trait like in node wallet: https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedKey.scala#L16

I agree. I don't see `ExtendedKey` trait being used in the code, but feel free to implement it if you find it useful.

ross-weir: > Good catch on child! I don't see a derive with full path in node wallet but it's up to you.

My Scala is bad but I was thinking that's what this was doing: https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedKey.scala#L37

Which is covered by these tests in the original link (https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKeySpec.scala#L37-L50)

greenhat: > > Good catch on child! I don't see a derive with full path in node wallet but it's up to you.
> 
> My Scala is bad but I was thinking that's what this was doing, or at least the difference in the paths: https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/main/scala/org/ergoplatform/wallet/secrets/ExtendedKey.scala#L37
> 
> Which is covered by these tests in the original link (https://github.com/ergoplatform/ergo/blob/c320810c498bca25a44197840c7c5a86440c5906/ergo-wallet/src/test/scala/org/ergoplatform/wallet/secrets/ExtendedSecretKeySpec.scala#L37-L50)

Oh yes, you're right, I missed it. I was looking if the trait is used as generic type somewhere to accommodate both.

ergo-lib/Cargo.toml
```diff
@@ -33,6 +33,7 @@ proptest-derive = {version = "0.3.0", optional = true }
 sha2 = { version = "0.9.8" }
 hmac = { version = "0.11.0" }
 pbkdf2 = "0.8"
+bs58 = "0.4.0"
```
greenhat, line 36: Are you sure it's needed? I did not find any usage in the code.

ross-weir: At the moment only used for comparing against node test vectors: https://github.com/ergoplatform/sigma-rust/pull/498/files#diff-214500aee5bfd9edaf034a03b5e268694ba68ef0bd0b3c79107ac39222b273c3R275

greenhat: Oh, I missed it. I was running grep on github's PR webpage. :)
Can you move it to dev-dependencies then, or do you think we're going to need it here soon anyway?

ross-weir: Moved to `dev-deps`, easy to move back to `deps` if it's ever needed 👍

greenhat (approved):
Great! Thank you!

Issue #481, opened 2021-11-19
Draft the ergo-p2p networking layer architecture


greenhat:
Requirements:
- runs on iOS/Android(async interop, callbacks);

Wasm is out of the question since TCP networking is not allowed in the browser VM.

Issue #495, opened 2021-12-01
Payment verification workflow with SPV


greenhat:
Using NiPoPoW described in [KMZ17](https://fc20.ifca.ai/preproceedings/74.pdf)

## Goal 
Prove that a payment (ERG and/or tokens) was made on the blockchain.

### Step 1.1. Prove that a block with a given id exists in the blockchain.

1.1.1. Ask proofs from known nodes via [`/nipopow/proof/{m}/{k}/{headerId}`](https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowAlgos.scala#L188-L271)

1.1.2. Create [`NipopowVerifier`](https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/local/NipopowVerifier.scala#L14-L28) Rust equvalent and feed all received proofs into `NipopowVerifier::process`
1.1.3. Check that `verifier.bestProof().suffixHead.id` is our block id;

### Step 1.2. Prove that a tx with a given id exists in the blockchain.

1.2.1. Ask any node for `MerkleProof` via [`/blocks/{headerId}/proofFor/{txId}`](https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/http/api/BlocksApiRoute.scala#L72-L85);
1.2.2. Confirm the tx id is in the block with [`merkleProof.valid(header.transactionsRoot)`](https://github.com/input-output-hk/scrypto/blob/2237785131ee0946e44d6a5aec2ef0fc0731efce/src/main/scala/scorex/crypto/authds/merkle/MerkleProof.scala#L28-L38);

### Step 2.1 Get the whole tx for tx id from Step 1 
It can either be provided along with the tx id in Step 1 or fetched from the node. It is verified by serializing the tx and calculating its id and then comparing to the tx id confirmed in Step 1.2.

### Step 2.2. Confirm our payment is in this tx
Find the output box protected with a certain script (e.g our PK) and check that its value and/or tokens are against expectations.

greenhat:
@kushti Could you please confirm the above workflow is correct? Also, I have the following questions:
[...]

PR #500, opened 2021-12-07
P2P networking layer draft

greenhat:
Ref #481

This PR partially drafts p2p networking layer. 
Missing parts (to be implemented later):
- exchange of the handshake messages;
- manage open TCP connections;

### DIFF
--- a/ergo-p2p/Cargo.toml
+++ b/ergo-p2p/Cargo.toml
@@ -14,13 +14,17 @@ exclude = [
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-
 sigma-ser = { version = "^0.3.0", path = "../sigma-ser" }
 sigma-util = { version = "^0.3.0", path = "../sigma-util" }
 thiserror = "1"
 derive_more = "0.99"
 bounded-vec = { version = "^0.5.0" }
 proptest-derive = {version = "0.3.0", optional = true }
+
+futures = "0.3"
+
+tracing = "0.1"
+tracing-error = { version = "0.1.2", features = ["traced-error"] }
 
 [dependencies.proptest]
 # wasm support, via https://altsysrq.github.io/proptest-book/proptest/wasm.html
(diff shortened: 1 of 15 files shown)


ergo-p2p/Cargo.toml
```diff
@@ -14,14 +14,18 @@ exclude = [
...
 [dependencies]
-
 sigma-ser = { version = "^0.3.0", path = "../sigma-ser" }
 sigma-util = { version = "^0.3.0", path = "../sigma-util" }
 thiserror = "1"
 derive_more = "0.99"
 bounded-vec = { version = "^0.5.0" }
 proptest-derive = {version = "0.3.0", optional = true }
 
+futures = "0.3"
```
ross-weir, line 24: Super early in the draft but just wanted to mention:

`tokio` might be a good choice for the networking/async layer if you haven't looked at it 😃

greenhat: Oh, there will be `tokio` in it for sure. :) Thanks for checking in!

greenhat:
I'm switching to REST-based peer messaging for the SPV client https://github.com/ergoplatform/sigma-rust/issues/495 and I want to merge it since I have no immediate plan to continue working on it.

ergo-p2p/src/client.rs
```diff
@@ -0,0 +1,29 @@
+#![allow(dead_code)] // TODO: remove
+
+use futures::channel::mpsc;
+use futures::channel::oneshot;
```
ross-weir, line 4: Out of curiosity, any particular reason we're going with `futures-rs` for these over `tokio`?

Had a bit of a google for comparisons but didn't find that much apart from `futures-rs` being developed by the `rust` team

ross-weir (approved):

ergo-p2p/src/client.rs
greenhat: Good catch! No reason apart from it was used in Zebra. Switched to tokio's.

PR #505, opened 2021-12-17
Add mnemonic generation

ross-weir:
Adds mnemonic phrase generation


ross-weir:
@greenhat Just wanted to check before adding bindings: should we put any of this behind a feature flag? The wordlists add a bit of bloat

greenhat:
Good idea! Yes, please put it behind the feature.

Best regards,
Denys Zadorozhnyi

ergo-lib/src/wallet/mnemonic.rs
```diff
@@ -1,8 +1,19 @@
 //! Mnemonic operations according to BIP32/BIP39
 
+#[cfg(feature = "mnemonic_gen")]
```
greenhat, line 3: I'd rather have `mnemonic_gen` feature attribute on the whole module, to avoid putting on almost every statement. `Mnemonic::to_seed` could be extracted into a function and moved out.

greenhat (changes requested):
Looking good! Please check my comments!

ross-weir:
> Looking good! Please check my comments!

Done

I also removed mnemonic gen from wasm, I couldn't see any good way to conditionally include it. Using a dedicated `bip39` package for this seems like a better/more flexible option

greenhat (approved):
Great! Thank you!

PR #508, opened 2021-12-18
Add address generation demo

ross-weir:
Port of the `AddressGenerationDemo` from node: https://github.com/ergoplatform/ergo/blob/master/ergo-wallet/src/test/java/org/ergoplatform/wallet/AddressGenerationDemo.java

Added some extra small methods to fill in gaps of missing functionality.


bindings/ergo-lib-wasm/src/wallet/derivation_path.rs
```diff
@@ -41,6 +41,30 @@ impl DerivationPath {
...
+    /// Returns a new path with the last element of the deriviation path being increased, e.g. m/1/2 -> m/1/3
+    /// Returns an empty path error if the path is empty (master node)
+    pub fn next(&self) -> Result<DerivationPath, JsValue> {
+        Ok(self.0.next().map_err(to_js)?.into())
+    }
+
+    /// String representation of derivation path
+    /// E.g m/44'/429'/0'/0/1
+    #[wasm_bindgen(js_name = toString)]
+    #[allow(clippy::inherent_to_string)]
```
ross-weir, line 63: I thought this was a better option for the `wasm` bindings, `toString` is used for displays in JS consoles/debuggers/etc, if I implemented `Display` trait we wouldn't get nice displays in JS land by default

bindings/ergo-lib-wasm/examples/address-generation-demo/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    let changePath = DerivationPath.new(0, new Uint32Array([0]));
+    const changeSecretKey = deriveSecretKey(rootSecret, changePath);
+    const changePubKey = changeSecretKey.public_key();
+    const changeAddress = NetworkAddress.new(NetworkPrefix.Mainnet, changePubKey.to_address());
+
+    console.log(`Change address: ${changeAddress.to_base58()}`);
+
+    // This is currently required becuase the line:
+    // `const changeSecretKey = deriveSecretKey(rootSecret, changePath);`
+    // Takes ownership of the changePath pointer and frees it so it's null when we get to this point
```
ross-weir, line 38: Maybe we need to use references a bit more in wasm bindings?

greenhat: Oh yes, I use the references almost always because of that. I try to catch them in review as well, but sometimes that get through.

bindings/ergo-lib-wasm/src/wallet/derivation_path.rs
greenhat: Good call!

greenhat (approved):
Looking great! Thank you!

bindings/ergo-lib-wasm/examples/address-generation-demo/index.ts
ross-weir: That's likely my bad, I'll be using these wasm bindings a heap so I'll fix them up as I come across them!

PR #511, opened 2021-12-21
exposing `Coll[Coll[Byte]]`,  `Coll[Int]` & `BigInt`

sahandzou:


sahandzou:
Hi, I implemented the missing Wasm bindings ready to review!

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -71,6 +72,13 @@ impl Constant {
             .map(I64::from)
     }
 
+    /// Create from byte array(BigInt)
+    pub fn from_byte_array_bigint(num: &[u8]) -> Constant {
```
greenhat, line 76: Let's say what format is expected here:
```suggestion
    pub fn from_bigint_signed_bytes_be(num: &[u8]) -> Constant {
```

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -71,6 +72,13 @@ impl Constant {
             .map(I64::from)
     }
 
+    /// Create from byte array(BigInt)
```
greenhat, line 75:
```suggestion
    /// Create BigInt constant from byte array (signed bytes bit-endian)
```

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -83,6 +91,46 @@ impl Constant {
             .map_err(to_js)
     }
 
+    /// Create `Coll[Int]` from string array
+    #[allow(clippy::boxed_local)]
+    pub fn from_i32_str_array(arr: Box<[JsValue]>) -> Result<Constant, JsValue> {
```
greenhat, line 96: Why use string and not JS number here? In i64 equivalent, the string was used since 64 bits do not fit in JS number.

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -123,6 +171,19 @@ impl Constant {
             .collect())
     }
 
+    /// Extract `Coll[Coll[Byte]]` as byte array
+    pub fn to_coll_coll_byte_array(&self) -> Result<Vec<Uint8Array>, JsValue> {
```
greenhat, line 175:
```suggestion
    pub fn to_coll_coll_byte(&self) -> Result<Vec<Uint8Array>, JsValue> {
```
`array` is redundant since `coll_coll` already conveys the type.

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -123,6 +171,19 @@ impl Constant {
             .collect())
     }
 
+    /// Extract `Coll[Coll[Byte]]` as byte array
```
greenhat, line 174:
```suggestion
    /// Extract `Coll[Coll[Byte]]` as array of byte arrays
```

greenhat (changes requested):
Looking good! Please see my notes.

greenhat:
Just a note. If you're working on a PR that is not yet ready for review, you can start with a draft PR and make it ready when it's ready for review.

bindings/ergo-lib-wasm/src/ast.rs
sahandzou: Yes, Good point I rewrite the function.

sahandzou:
> Just a note. If you're working on a PR that is not yet ready for review, you can start with a draft PR and make it ready when it's ready for review.

Yes, that was my mistake. I make the changes with respect to your comments and also write the Js test for the `i32` array round trip in the last commit.

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -83,6 +91,26 @@ impl Constant {
             .map_err(to_js)
     }
 
+    /// Create `Coll[Int]` from string array
```
greenhat, line 94:
```suggestion
    /// Create `Coll[Int]` from integer array
```

bindings/ergo-lib-wasm/src/ast.rs
```diff
@@ -83,6 +91,26 @@ impl Constant {
...
+        arr.iter()
+            .try_fold(vec![], |mut acc, l| {
+                acc.push(*l);
+                Ok(acc)
+            })
+            .map(|longs| longs.into())
+            .map(Constant)
+    }
+
+    /// Extract `Coll[Int]` as string array
```
greenhat, line 106:
```suggestion
    /// Extract `Coll[Int]` as integer array
```

sahandzou: I merged new comments.

greenhat (approved):
Great! Thank you!

sahandzou:
> Great! Thank you!

Thank you too!

Issue #510, opened 2021-12-20
Draft REST API networking layer for SPV client API


greenhat:
SPV workflow is described in #495 

Todo:
- peer discovery #544;
- request NiPoPoW proof from multiple peers;

Requirements:
- Wasm;
- JNI/Android;
- Swift/iOS;

greenhat:
Closed in favor #544, #545

PR #513, opened 2021-12-21
Draft the REST API for SPV client for Rust <-> C FFI <-> Swift and Wasm

greenhat:
Ref #510 

This PR drafts the async API for long-running tasks for the SPV node API with timeout and task cancellation support.

### Wasm
Rust futures exposed as JS futures and underlying `reqwest` implementation for Wasm target using browser's Fetch API.

### C/Swift
Swift API is built on passing user's closures as callbacks to be called on task completion. 
The call flow with under-the-hood implementation looks like this:
- Wrap user's closure to convert the result from C types to Swift types and increment its reference count. See [WrapClosure.swift](https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-ios/Sources/ErgoLib/WrapClosure.swift#L12-L57)
- In `ergo-lib-c-core` make a Rust Future calling `ergo-rest` `async` API and wrap it in `Abortable` returning abort handle. See [abortable.rs](https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-c-core/src/rest/api/node/abortable.rs#L7-L19) 
- In case of task cancellation (called `RequestHandle::abort`) besides aborting the task calls for user's closure release on the Swift side via [`CompletionCallback::callback_release`](https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-ios/Sources/ErgoLib/WrapClosure.swift#L46-L53)

From the perspective of extending the C/Swift API with new tasks the developer's workflow should be the following:
- Add `ergo-lib-c-core` function similar to 
https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-c-core/src/rest/api/node.rs#L17-L35
- Wrap it in Swift similar to https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-ios/Sources/ErgoLib/RestNodeApi.swift#L20-L32

### Misc
This PR moves some types from ergotree-ir and ergo-lib crates to ergo-chain-types crate.


greenhat:
Hey @ross-weir @kettlebell
I'd like to get your feedback on two different approaches in Swift API for long-running (network) request(s). I started with the idea of exposing async Rust and it became ugly very quickly and I want your opinion on whether we should not bother with an async FFI at all.

Exhibit A. Async Rust to async Swift:
Swift - 
https://github.com/ergoplatform/sigma-rust/blob/95e533beb41ce3667766a1a869e11cc6ff3e1850/bindings/ergo-lib-ios/Sources/ErgoLib/RestNodeApiAsync.swift#L12-L66
Rust counterpart - 
https://github.com/ergoplatform/sigma-rust/blob/95e533beb41ce3667766a1a869e11cc6ff3e1850/bindings/ergo-lib-c-core/src/rest/api/node_async.rs#L14-L74

Besides high complexity and poor ability to reason about this code, it also lacks more granular thread control for Swift dev (think calling callbacks on UI thread).

And here is a "dumbed down" version - Exhibit B. Async Rust to blocking Swift:
Swift - 
https://github.com/ergoplatform/sigma-rust/blob/95e533beb41ce3667766a1a869e11cc6ff3e1850/bindings/ergo-lib-ios/Sources/ErgoLib/RestNodeApi.swift#L6-L12

Rust counterpart - 
https://github.com/ergoplatform/sigma-rust/blob/95e533beb41ce3667766a1a869e11cc6ff3e1850/bindings/ergo-lib-c-core/src/rest/api/node.rs#L6-L25

It is obviously lacking async Swift API but we can either wrap it with GCD on the Swift side or we can leave it as is (blocking) and a Swift dev will wrap it in GCD/NSOperation/whatever on their own terms in their app.


Comparing Exhibit A and B I really want to ditch A completely and go with B only. Do you see any benefits in going with A?

kettlebell:
With approach A, we can go further with it and provide a modern async API on the swift side. The latest swift version (5.5) provides async/await functionality that's comparable to what's in rust, with the restriction that iOS must be at least version 13 (I think it covers most Apple devices). This article describes how to get a modern async interface from completion handlers: https://www.hackingwithswift.com/articles/233/whats-new-in-swift-5-5 (scroll down to "Continuations for interfacing async tasks with synchronous code" section.)

The implementation isn't pleasant as you've demonstrated, but the resulting swift API could look very much like the WASM one.

With approach B, I wonder how you could get task cancellation with blocking calls? And spawning a new runtime instance for each REST call, we could end up spawning heaps of threads (might be an issue for iOS?).

greenhat:
> With approach A, we can go further with it and provide a modern async API on the swift side. The latest swift version (5.5) provides async/await functionality that's comparable to what's in rust, with the restriction that iOS must be at least version 13 (I think it covers most Apple devices). This article describes how to get a modern async interface from completion handlers: https://www.hackingwithswift.com/articles/233/whats-new-in-swift-5-5 (scroll down to "Continuations for interfacing async tasks with synchronous code" section.)
 
Whoa! Swift has async/await now! Nice! I totally missed it. 
OTOH, one more layer on top of async Rust/Swift "onion" might not look good. I wonder if a blocking FFI (blocking Rust/async Swift) can be wrapped in async/await as well.

> The implementation isn't pleasant as you've demonstrated, but the resulting swift API could look very much like the rust/WASM one.
> 
> With approach B, I wonder how you could get task cancellation with blocking calls? 

Good question! I missed the cancellation completely. Well, it seems that canceling an async Rust means dropping a Future. One way of doing it is to wrap it with `select` along with "signaling" Future that ends on a signal and then somehow send a signal from Swift. Damn it, how deep is this rabbit hole? :) I wonder if blocking call (approach B) wrapped in async Swift can be canceled more easily.

> And spawning a new runtime instance for each REST call, we could end up spawning heaps of threads (might be an issue for iOS?).

Good catch! I was thinking of using a "lighter" `current_thread` runtime but creating it on every request is probably an overkill too. We can store the wrapped runtime in `RestNodeApi` class (like in `RustNodeApiAsync`) and pass it into Rust.

ross-weir:
I like keeping it simple to start with option B for bindings/FFI and leaving it up to consumers to wrap with async stuff in their language of choice.

> Good question! I missed the cancellation completely.

Do sync APIs normally have cancellation? If we go the sync route could we allow library consumers to supply a `timeout` config for ops, then if they want to wrap it in async stuff they can handle cancellation - looks like it's supported with GCD

kettlebell:
> Whoa! Swift has async/await now! Nice! I totally missed it. OTOH, one more layer on top of async Rust/Swift "onion" might not look good. I wonder if a blocking FFI (blocking Rust/async Swift) can be wrapped in async/await as well.

Another possible approach is to directly implement the REST API at the swift level. There's this popular library https://github.com/Alamofire/Alamofire (which has been recently updated for swift async/await) that's pretty much like `Reqwest` for swift. 

>  Well, it seems that canceling an async Rust means dropping a Future. One way of doing it is to wrap it with `select` along with "signaling" Future that ends on a signal and then somehow send a signal from Swift. Damn it, how deep is this rabbit hole? :)

There's already a pre-made solution from the `futures` crate: https://docs.rs/futures-util/0.3.19/futures_util/future/struct.Abortable.html. I've combined it with a `oneshot` channel and it works well. We can easily wrap up the `tx` side of it within FFI and trigger it from Swift to effect the cancellation.

greenhat:
> I like keeping it simple to start with option B for bindings/FFI and leaving it up to consumers to wrap with async stuff in their language of choice.
> 
> > Good question! I missed the cancellation completely.
> 
> Do sync APIs normally have cancellation? If we go the sync route could we allow library consumers to supply a `timeout` config for ops, then if they want to wrap it in async stuff they can handle cancellation - looks like it's supported with GCD

I don't think sync API can have a cancellation. I mean how to kill a blocking call besides maybe killing the whole thread it runs on.

greenhat:
Hey, @kettlebell thanks for pushing "true" async approach, and thank you @ross-weir for advocating for simplicity! I think I managed to implement one without sacrificing the other. With both timeout and task cancellation implemented I'd like to merge this PR and continue to add more tasks to the API (peer discovery, etc.) in the future PRs. I'd appreciate it if you take a look at the implementation details I highlighted in the PR's description.

bindings/ergo-lib-c-core/src/rest/api/node.rs
```diff
@@ -0,0 +1,35 @@
...
+) -> Result<(), Error> {
+    let runtime = const_ptr_as_ref(runtime_ptr, "runtime_ptr")?;
+    let node_conf = const_ptr_as_ref(node_conf_ptr, "node_conf_ptr")?.0;
+    let release_callback: ReleaseCallbackWrapper = (&callback).into();
+    let abort_handle = spawn_abortable(runtime, async move {
+        match ergo_lib::ergo_rest::api::node::get_info(node_conf).await {
+            Ok(node_info) => callback.succeeded(node_info),
+            Err(e) => callback.failed(e.into()),
+        }
+    })?;
```
kettlebell, line 31: Really straightforward to wrap!

bindings/ergo-lib-c/cbindgen.toml
```diff
@@ -141,3 +141,11 @@ crates = ["ergo-lib-c"]
 all_features = false
 default_features = true
 features = []
+
+
+[ptr]
+# An optional string to decorate all pointers that are
+# required to be non null. Nullability is inferred from the Rust type: `&T`,
+# `&mut T` and `NonNull<T>` all require a valid pointer value.
+non_null_attribute = "_Nonnull"
```
kettlebell, line 150: I looked up this feature, very nice. I wonder if it's worth refactoring the rest of the codebase to use `NonNull`.

bindings/ergo-lib-ios/Sources/ErgoLib/NodeInfo.swift
```diff
@@ -0,0 +1,30 @@
...
+class NodeInfo: FromRawPtr {
+    internal var pointer: NodeInfoPtr
+
+    internal init(withRawPointer ptr: NodeInfoPtr) {
+        self.pointer = ptr
+    }
+
+    static func fromRawPtr(ptr: UnsafeRawPointer) -> Self {
+        return NodeInfo(withRawPointer: OpaquePointer(ptr)) as! Self
+    }
```
kettlebell, line 13: Any benefit to having `fromRawPtr` over the constructor?

bindings/ergo-lib-ios/Sources/ErgoLib/RestNodeApi.swift
```diff
@@ -0,0 +1,47 @@
...
+
+
+    // need macOS 12.0 on GA
+    /* func getInfoAsync(nodeConf: NodeConf) async throws -> NodeInfo { */
+    /*     try await withCheckedThrowingContinuation { continuation in */
+    /*         getInfo(nodeConf: nodeConf) { result in */ 
+    /*             continuation.resume(with: result) */
+    /*         } */
+    /*     } */
+    /* } */
```
kettlebell, line 42: It's nice that we can simply reuse `getInfo`.

bindings/ergo-lib-wasm/tests/test_rest_api.js
```diff
@@ -0,0 +1,13 @@
...
+  get_info, NodeConf,
+} from '../pkg/ergo_lib_wasm';
+
+// it('node REST API get_info endpoint', async () => {
+//   let node_conf = new NodeConf("213.239.193.208:9053");
+//   assert(node_conf != null);
+//   let res = await get_info(node_conf);
+//   assert(res != null);
+//   assert(node_conf != null);
+// });
```
kettlebell, line 13: Can we re-enable the test?

bindings/ergo-lib-c-core/src/rest/api/callback.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use ergo_lib::ergo_rest::NodeResponse;
+
+use crate::Error;
+
+/// Callback info from the Swift side
+#[repr(C)]
+pub struct CompletionCallback {
+    /// Wrapped Swift closure pointer (see WrapClosure.swift)
+    swift_closure: NonNull<c_void>,
```
ross-weir, line 13: If these are `swift` only should we put it behind a feature flag so it's not included in C?

ross-weir:
> Hey, @kettlebell thanks for pushing "true" async approach, and thank you @ross-weir for advocating for simplicity! I think I managed to implement one without sacrificing the other. With both timeout and task cancellation implemented I'd like to merge this PR and continue to add more tasks to the API (peer discovery, etc.) in the future PRs. I'd appreciate it if you take a look at the implementation details I highlighted in the PR's description.

Looking awesome, great work!

bindings/ergo-lib-c-core/src/rest/api/node.rs
greenhat: Thanks!

bindings/ergo-lib-ios/Sources/ErgoLib/RestNodeApi.swift
greenhat: Yeah, but I'm not sure that cancellation is possible in async/await. I need to dig into that later on.

bindings/ergo-lib-ios/Tests/ErgoLibTests/RestNodeApiTests.swift
```diff
@@ -0,0 +1,49 @@
...
+    /*     XCTAssertNoThrow(Task(priority: .medium) { */
+    /*         let nodeInfo = try await restNodeApi.getInfoAsync(nodeConf: nodeConf) */
+    /*         XCTAssert(!nodeInfo.getName().isEmpty) */
+    /*     }) */
+    /*     // test of re-using of tokio runtime */
+    /*     XCTAssertNoThrow(Task(priority: .medium) { */
+    /*         let nodeInfo = try await restNodeApi.getInfoAsync(nodeConf: nodeConf) */
+    /*         XCTAssert(!nodeInfo.getName().isEmpty) */
+    /*     }) */
+    /* } */
```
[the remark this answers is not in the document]
greenhat: Yeah, I totally agree. They did a great job implementing it in Swift.

bindings/ergo-lib-ios/Sources/ErgoLib/NodeInfo.swift
greenhat: Oh, it's an implementation of `FromRawPtr` protocol I made for `wrapClosure` - https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-ios/Sources/ErgoLib/WrapClosure.swift#L60 to get wrapped Swift response type out of Rust's pointer.

bindings/ergo-lib-wasm/tests/test_rest_api.js
greenhat: Good catch! But no, :) it does not work since we're running these JS test on node in CI and this implementation works only in the browser (`reqwest` implementation for Wasm target use browser's Fetch API). I made https://github.com/ergoplatform/sigma-rust/issues/522 to run JS tests in the browser on CI. When its implemented I hope I remember to uncomment it. :)

bindings/ergo-lib-c/cbindgen.toml
greenhat: Yes, I found it when I was constructing `CompletionCallback` in Swift - https://github.com/ergoplatform/sigma-rust/blob/92fb1e3a3b7e440734bb08b7b1f1502c8fb646b3/bindings/ergo-lib-c-core/src/rest/api/callback.rs#L9-L19
I believe it makes sense mainly in the struct properties. Do we have any, besides the new `*Callback`?

bindings/ergo-lib-c-core/src/rest/api/callback.rs
greenhat: Good question! It got me thinking that there is actually nothing inherently specific to Swift. So I renamed both callback structs and their fields and it made names and comments so much clearer! Thanks!
Here is how they look now: https://github.com/ergoplatform/sigma-rust/blob/500ad8b867ee3c882fec7ccccee5993260b05c49/bindings/ergo-lib-c-core/src/rest/api/callback.rs#L9-L20

greenhat:
@kettlebell @ross-weir Thank you for your reviews! I made some changes and addressed all the comments.

bindings/ergo-lib-c/cbindgen.toml
kettlebell: I just checked and the only places we have it are for `Return[Num|Bool|Option]`. These structs have the `error :: ErrorPtr` field which is supposed to be `null` if there's a valid value. So looks like we don't need to make further changes elsewhere.

PR #521, opened 2021-12-28
Add ergo-merkle-tree

SethDusek:
Hi, here is my first draft for ergo-merkle-tree, I've not added C/Swift bindings yet since there's some design decisions I'd like to get ironed out/approved beforehand.

Some thoughts:

- Right now, the only accepted hash for MerkleProof is 32 bytes long, same with leaf data. Trying to create a MerkleProof with any other size returns an error.  Should this instead be dynamic to accommodate any size hash/data?

- MerkleProof.leaf_data could be a borrowed slice, but I decided to make it an owned array since that makes it a bit easier to write bindings for since there's no lifetimes to deal with, another way would also be to not have the leaf data be part of the proof at all but instead have MerkleProof::valid's signature be:
```rust 
fn valid(&self, leaf_data: &[u8], expected_root: &[u8]) -> bool
```
but this approach would have a disadvantage since we wouldn't be able to directly deserialize the MerkleProof from JSON (see: tests/block_proof)

- Finally, I've added a prefixed_hash function to lib.rs, do you think it would be a better fit for the sigma-util crate instead (If there is some utility for it in other parts of ergo)?


greenhat:
> Hi, here is my first draft for ergo-merkle-tree, I've not added C/Swift bindings yet since there's some design decisions I'd like to get ironed out/approved beforehand.

Great! Thank you!

> 
> Some thoughts:
> 
>     * Right now, the only accepted hash for MerkleProof is 32 bytes long, same with leaf data. Trying to create a MerkleProof with any other size returns an error.  Should this instead be dynamic to accommodate any size hash/data?

I think it's safe to have it fixed to 32 bytes. 
> 
>     * MerkleProof.leaf_data could be a borrowed slice, but I decided to make it an owned array since that makes it a bit easier to write bindings for since there's no lifetimes to deal with, another way would also be to not have the leaf data be part of the proof at all but instead have MerkleProof::valid's signature be:
> 
> 
> ```rust
> fn valid(&self, leaf_data: &[u8], expected_root: &[u8]) -> bool
> ```
> 
> but this approach would have a disadvantage since we wouldn't be able to directly deserialize the MerkleProof from JSON (see: tests/block_proof)

Let's keep it as an owned array. Good call!

> 
>     * Finally, I've added a prefixed_hash function to lib.rs, do you think it would be a better fit for the sigma-util crate instead (If there is some utility for it in other parts of ergo)?

Good idea! Although I don't see it's used much in ergo node let's have it besides `blake2b256_hash` in `sigma_util`. Please, name it similarly for consistency.

bindings/ergo-lib-wasm/src/merkleproof.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// A level node in a merkle proof
+#[wasm_bindgen]
+#[derive(Serialize, Deserialize, Debug, Copy, Clone)]
+pub struct LevelNode(ergo_merkle_tree::LevelNode);
+
+#[wasm_bindgen]
+impl LevelNode {
+    /// Creates a new LevelNode from a 32 byte hash and side that the node belongs on in the tree. Fails if the digest is not 32 bytes
+    pub fn new(hash: &[u8], side: NodeSide) -> Result<LevelNode, JsValue> {
```
greenhat, line 46: Passed `side` becomes null in JS after this call. It's better to pass by reference. The same goes to `MerkleProof::add_node` below.

bindings/ergo-lib-c-core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ crate-type = ["lib"]
 base16 = "0.2.1"
 derive_more = "0.99"
 ergo-lib = { version = "^0.14.0", path = "../../ergo-lib" }
+ergo-merkle-tree = { version = "0.1.0", path = "../../ergo-merkle-tree" }
```
greenhat, line 17: To keep `bindings/ergo-lib-*` crates dependencies minimal I'd rather have it as a re-export in `ergo-lib` along with other dependencies - https://github.com/ergoplatform/sigma-rust/blob/95e533beb41ce3667766a1a869e11cc6ff3e1850/ergo-lib/src/lib.rs#L27-L34
The same goes for `ergo-lib-wasm`.

greenhat (changes requested):
It looks great! I appreciate JS and Swift tests! See my minor comments.

bindings/ergo-lib-wasm/src/merkleproof.rs
SethDusek: wasm_bindgen doesn't seem to have an implementation of RefFromWasmAbi for #[repr(u8)] enums. As an alternative, I've decided to pass a u8 instead, since NodeSide implements TryFrom<u8>. I'm not too familiar with wasm_bindgen, so let me know if there is a better way!

greenhat (approved):
Great! Thank you!

PR #530, opened 2022-01-10
Impl P2P integration testing suite(#480)

### LINKED ISSUE #480: [300 SigmaUSD] Integration testing suite for P2P networking
### Motivation
- To test our implementation of P2P networking against the reference implementation in Scala.
- To catch regressions introduced during the development.

### Implementation options
The idea is to test P2P requests and responses against running ergo node built from the master branch https://github.com/ergoplatform/ergo
In our tests, we should connect to the node and run our scenarios. Ideally, we want to make tests stateless, i.e. tests should not depend on changes to the node's state made in other tests. Although we can ignore it for now due to initial ergo-p2p requests being intended for the SPV node and does not change the node's state. In the future when we implement P2P requests that change the node's state we'd want them to be stateless. So it makes sense to lay the foundation for it now.

Possible options to run ergo node:
### 1. Launch as a standalone application
Launch the node as usual, let it sync with the mainnet, and then kill it after our tests are done.
Cons:
- wasting mainnet nodes resources;
- risk of banning by github since running nodes are against github actions TOS;

### 2. Run node's integration test
Create a special integration test that starts the node in its own net in mining (block generation) mode. This test can be then called via `sbt it:test [name]` command.

See the example of ergo integration test - https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/it/scala/org/ergoplatform/it/NodeRecoverySpec.scala#L14-L53

Clearly, we want the second option, i.e. run the node via its own integration test. Also, it fits into the future stateless tests requirement, where we just launch the node via its own integration test for any of our single tests.

### Todo
- Write an integration test in ergo node to run it in our tests;
- Establish GA workflow for p2p integration tests;

kettlebell:
Close #480


greenhat (approved):
Great job! Thank you!

PR #531, opened 2022-01-11
[WIP] Sigma `AtLeast` tests

### LINKED ISSUE #248: [300 SigmaUSD] Sigma protocol implementation for AtLeast

ross-weir:
closes #248

### DIFF
--- a/ergotree-interpreter/src/sigma_protocol/prover.rs
+++ b/ergotree-interpreter/src/sigma_protocol/prover.rs
@@ -843,7 +843,7 @@ fn step9_real_threshold(ct: CthresholdUnproven) -> Result<Option<ProofTree>, Pro
                 ProofTree::UnprovenTree(unpt) => unpt.challenge(),
             };
             if let Some(challenge) = challenge_opt {
-                points.append(&mut one_based_idx.to_be_bytes().to_vec());
+                points.push(one_based_idx as u8);
                 values.push(challenge.into());
             };
         }


kettlebell:
Hi @ross-weir, I got your tests to pass. The root of the problem was a bug in my code, which is fixed in https://github.com/ergoplatform/sigma-rust/pull/534. After that you just need to revert some code, see my comments.

ergotree-interpreter/src/sigma_protocol/fiat_shamir.rs
```diff
@@ -133,6 +135,36 @@ fn fiat_shamir_write_bytes<W: SigmaByteWrite>(
         ProofTreeKind::Conjecture(c) => {
             w.put_u8(INTERNAL_NODE_PREFIX)?;
             w.put_u8(c.conjecture_type() as u8)?;
+            match tree {
```
kettlebell, line 138: Remove the match statement.

ergotree-interpreter/src/sigma_protocol/prover.rs
```diff
@@ -843,7 +843,7 @@ fn step9_real_threshold(ct: CthresholdUnproven) -> Result<Option<ProofTree>, Pro
                 ProofTree::UnprovenTree(unpt) => unpt.challenge(),
             };
             if let Some(challenge) = challenge_opt {
-                points.append(&mut one_based_idx.to_be_bytes().to_vec());
```
kettlebell, line 846: Revert this to the original line.

ross-weir: Original line causes the following error:

`Gf2_192PolyError(InterpolatePointsAndValuesLengthDiffer)`

ergotree-interpreter/src/sigma_protocol/fiat_shamir.rs
ross-weir: Just to confirm, removing the match makes our serialization different to sigma-state - that's correct?

kettlebell: Ah, I can't see what happened since the force-push. I forked your PR as of last night (~ 9 hours ago), made these changes, and everything passed.

The match statement I highlighted contained a match arm which had a call to the writer: `w.put_u8(&k)`. The bug I fixed in my code mistakenly emitted 24 extra bytes.

kettlebell: (edited previous comment showing what was failing on my side)

ross-weir: Damn that sucks, without the `w.put_u8(*k)` included in serialization the signing spec test vector `sig_test_vector_threshold` from `sigma-state` fails. 🤔 

If you give it another shot with the current branch `sigma_protocol::verifier::tests::test_prover_verifier_atleast` will fail, it starts failing when there's [more than 1 proposition threshold bound](https://github.com/ergoplatform/sigma-rust/pull/531/files#diff-5a1347bdbfcb71876cb34a1152319990ca71c461485c5106087a5ac8b3962c3dR406). Haven't had any success figuring it out yet

ergotree-interpreter/src/sigma_protocol/prover.rs
kettlebell: Your change here is correct. Each point has to be a distinct `u8`.

ergotree-interpreter/src/sigma_protocol/fiat_shamir.rs
kettlebell: Yeah disregard my suggestion to get rid of that line (you definitely know more than me about this section of the code).

I'll check it out.

kettlebell: Aha, got it! Change the `==` to `<=` for this line: https://github.com/ross-weir/sigma-rust/blob/4641814fb535656b72f360e88ea445df7a26d30c/ergotree-ir/src/sigma_protocol/sigma_boolean/cthreshold.rs#L50

ross-weir: Thanks so much for the help with this!

Although it makes this test pass it makes a different one fail and the comments in `sigma-state` specify we convert to `AND` if `cur_k == children_left` so I'm not sure this is the way to go

PR #534, opened 2022-01-11
Fix bugs with `Gf2_192Poly::to_bytes`

### LINKED ISSUE #533: Bugs with `Gf2_192Poly::to_bytes`
The original Scala implementation for this method (https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/main/java/gf2t/GF2_192_Poly.java#L214-L224) gives the option to serialize a Gf2 polynomial with and without the zero-degree coefficient.

My adaptation of this method assumed that the serialization would include the zero-degree coefficient, but that is not how it's used in `ergotree-interpreter`: https://github.com/ergoplatform/sigma-rust/blob/c6919c6a248695efd6ff3febc0ea8f30a871786b/ergotree-interpreter/src/sigma_protocol/sig_serializer.rs#L96-L97. Since this is the only place in the repo where `to_bytes` is used, we can make it not emit the zero-degree coefficient.

Also I didn't account for the `<=deg` within the for loop in the Scala code.

kettlebell:
Close #533

### DIFF
--- a/ergotree-interpreter/src/sigma_protocol/sig_serializer.rs
+++ b/ergotree-interpreter/src/sigma_protocol/sig_serializer.rs
@@ -93,7 +93,7 @@ fn sig_write_bytes<W: SigmaByteWrite>(
                 k: _,
                 polynomial,
             } => {
-                // write the polynomial, except the zero coefficient
+                // write the polynomial, except the zero-degree coefficient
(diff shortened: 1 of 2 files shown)


greenhat (approved):
Looking great! Thank you for taking care of this!

PR #535, opened 2022-01-12
Signing threshold test vector

ross-weir:
Completes:

`test verifier with proofs from mainnet/sigmastate Scala test vector;`

in #248 

Splitting this out from my other PR as its turning into a bit of a rabbit hole.


greenhat (approved):
Looking great! Thank you!

PR #539, opened 2022-01-22
Add prove -> verify `atleast` test

### LINKED ISSUE #248: [300 SigmaUSD] Sigma protocol implementation for AtLeast

ross-weir:
closes #248 

Addresses the task in the issue:

`test prover with verifier by generating arbitrary AtLeast IR node proving it and then passing through the verifier (see verifier::tests for examples);`

### DIFF
--- a/ergotree-interpreter/src/sigma_protocol/verifier.rs
+++ b/ergotree-interpreter/src/sigma_protocol/verifier.rs
@@ -115,8 +115,6 @@ pub fn verify_signature(
 fn check_commitments(sp: UncheckedTree, message: &[u8]) -> Result<bool, VerifierError> {
     // Perform Verifier Step 4
     let new_root = compute_commitments(sp);
-    println!("new_root challenge {:?}", new_root.challenge());
-
     let mut s = fiat_shamir_tree_to_bytes(&new_root.clone().into())?;
     s.append(&mut message.to_vec());
     // Verifier Steps 5-6: Convert the tree to a string `s` for input to the Fiat-Shamir hash function,
@@ -124,9 +122,6 @@ fn check_commitments(sp: UncheckedTree, message: &[u8]) -> Result<bool, Verifier
     // Accept the proof if the challenge at the root of the tree is equal to the Fiat-Shamir hash of `s`
     // (and, if applicable,  the associated data). Reject otherwise.
     let expected_challenge = fiat_shamir_hash_fn(s.as_slice());
-
-    println!("expected_challenge {:?}", expected_challenge);
-
     Ok(new_root.challenge() == expected_challenge.into())
 }
 
--- a/ergotree-interpreter/src/sigma_protocol/verifier.rs
+++ b/ergotree-interpreter/src/sigma_protocol/verifier.rs
@@ -437,7 +437,7 @@ mod tests {
             );
             let input = Constant {
                 tpe: SType::SColl(SType::SSigmaProp.into()),
-                v: inputs.clone(),
+                v: inputs,
             }
             .into();
             let expr: Expr = Atleast::new(bound, input).unwrap().into();
--- a/ergotree-interpreter/src/sigma_protocol/verifier.rs
+++ b/ergotree-interpreter/src/sigma_protocol/verifier.rs
@@ -115,6 +115,8 @@ pub fn verify_signature(
 fn check_commitments(sp: UncheckedTree, message: &[u8]) -> Result<bool, VerifierError> {
     // Perform Verifier Step 4
     let new_root = compute_commitments(sp);
+    println!("new_root challenge {:?}", new_root.challenge());
+
     let mut s = fiat_shamir_tree_to_bytes(&new_root.clone().into())?;
     s.append(&mut message.to_vec());
     // Verifier Steps 5-6: Convert the tree to a string `s` for input to the Fiat-Shamir hash function,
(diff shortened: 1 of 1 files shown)


ergotree-interpreter/src/sigma_protocol/verifier.rs
```diff
@@ -392,5 +402,48 @@ mod tests {
                 prop_assert_eq!(ver_res.unwrap().result, true, "verify failed on secret: {:?}", &secret);
             }
         }
+
+        #[test]
+        fn test_prover_verifier_atleast(secret1 in any::<DlogProverInput>(),
```
ross-weir, line 407: @greenhat Does this test look OK? If the implementation was sound then it should pass?

Still having troubles figuring out why verification fails

greenhat: The test looks fine. Yes, it should.
There is a test with test vectors for `AtLeast` in Scala - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/795c176debfe105646a7deef3f01bd537b79c81d/sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala#L119-L135
Could you implement it and then we'll see if our verifier is correct and the problem is in the prover.
We have corresponding tests for similar test vectors already in https://github.com/ergoplatform/sigma-rust/blob/9b5ca5a9e4ccbc076991c9db349645c4db848cae/ergotree-interpreter/tests/signing_spec_tests.rs#L325-L393

ross-weir: I implemented this in a previous PR: https://github.com/ergoplatform/sigma-rust/blob/c308476ad074147596b03fb9f3dcb04018823a72/ergotree-interpreter/tests/signing_spec_tests.rs#L400-L466

I totally blanked on this, so that narrows it down to the prover

greenhat: Sorry, I missed it. 
I feel you. Hunting for the error in the prover sucks. I appreciate you're taking care of this!

greenhat:
I found some differences in the prover - https://github.com/ergoplatform/sigma-rust/pull/548 but they do not fix the AtLeast code path.
I wonder if mocking the randomness with hard-coded values and running the Scala and Rust provers with the same predefined set of keys with debug prints sprinkled everywhere is the easiest method to find the bug?

ross-weir:
> I found some differences in the prover - https://github.com/ergoplatform/sigma-rust/pull/548 but they do not fix the AtLeast code path.
> I wonder if mocking the randomness with hard-coded values and running the Scala and Rust provers with the same predefined set of keys with debug prints sprinkled everywhere is the easiest method to find the bug?

Good idea, I'll give mocking a shot!

ross-weir:
On step 7 of the prover (`fiat_shamir_tree_to_bytes`), the calculated bytes are different. The buffer specifically starts to differ at offset 50

In `fiat_shamir_write_bytes` things appear to start to differ in `ProofTreeKind::Leaf` match arm, the prop bytes are the same as `sigma-state` but commitment bytes differ.

Going a bit deeper, `EcPoint.sigma_serialize` is generating different results compared to `GroupElementSerializer.serialize` in `sigma-state` when getting the commitment bytes

@greenhat any of that ☝️ ring any bells?

greenhat:
Thank you for getting back to this! I appreciate it.

> On step 7 of the prover (`fiat_shamir_tree_to_bytes`), the calculated bytes are different. The buffer specifically starts to differ at offset 50
> 
> In `fiat_shamir_write_bytes` things appear to start to differ in `ProofTreeKind::Leaf` match arm, the prop bytes are the same as `sigma-state` but commitment bytes differ.

I think it's due to the randomness involved in the process of generating the commitments. Or did you mocked the sources of randomness?

> 
> Going a bit deeper, `EcPoint.sigma_serialize` is generating different results compared to `GroupElementSerializer.serialize` in `sigma-state` when getting the commitment bytes

Are `EcPoint`s the same?

ross-weir:
> Thank you for getting back to this! I appreciate it.
> 
> > On step 7 of the prover (`fiat_shamir_tree_to_bytes`), the calculated bytes are different. The buffer specifically starts to differ at offset 50
> > In `fiat_shamir_write_bytes` things appear to start to differ in `ProofTreeKind::Leaf` match arm, the prop bytes are the same as `sigma-state` but commitment bytes differ.
> 
> I think it's due to the randomness involved in the process of generating the commitments. Or did you mock the sources of randomness?
> 
> > Going a bit deeper, `EcPoint.sigma_serialize` is generating different results compared to `GroupElementSerializer.serialize` in `sigma-state` when getting the commitment bytes
> 
> Are `EcPoint`s the same?

Good point.

Small update:
I tried again hard coding the commitments with a known good set from `sigma-state` and the test passes. Taking a deeper look into `EcPoint`s

greenhat:
There was a fix recently to `CthresholdUnproven` in the prover - https://github.com/ergoplatform/sigma-rust/pull/598/files#r921832411
Do you think it might be it?

ross-weir:
> There was a fix recently to `CthresholdUnproven` in the prover - https://github.com/ergoplatform/sigma-rust/pull/598/files#r921832411 Do you think it might be it?

Yep, that appears to have been the issue. Nice fix guys!

greenhat (approved):
Looking great! Thank you!

PR #547, opened 2022-02-08
fix DlogProverInput::from_biguint, add check in ExtSecretKey derivation;

greenhat:
During the bug hunt in https://github.com/ergoplatform/ergo/issues/1627 I found and fixed the following:
- missing check in our BIP32 secret key derivation;
- less than 32 bit bigint import fail in `DlogProverInput::from_biguint`;

Also added test vectors from ergo-wallet incorrect BIP32 secret derivation and appkit test vectors.


ross-weir (approved):
Awesome work investigating the derivation issues. 👍 

My bad on missing the zero case, glad that was found and fixed!

greenhat:
Thank you! And no worries! I believe zero case probability should be very very low. :)

PR #551, opened 2022-03-02
Implement PoPowHeader type

### LINKED ISSUE #489: [100 SigmaUSD] Create PoPowHeader type
In ergo-nipopow crate. Implement binary (ScorexSerializable trait) and JSON serialization. Use `Header` type from ergotree-ir (or from ergo-chain-types if #488 is implemented).
Scala reference implementation - https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/modifiers/history/popow/PoPowHeader.scala#L13-L83

SethDusek:
Closes #489

### DIFF
--- a/bindings/ergo-lib-c-core/src/nipopow.rs
+++ b/bindings/ergo-lib-c-core/src/nipopow.rs
@@ -1,7 +1,7 @@
 //! Bindings for NiPoPow
 
 use crate::{
-    block_header::{BlockHeader, ConstBlockIdPtr},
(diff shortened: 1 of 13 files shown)


SethDusek:
Should be good to go now, I've added BlockIds collection to Swift too assuming that an end-user wants to access the interlinks for PoPowHeader. I'm not sure what to add in the way of unit testing for PoPowHeader though, I've added a JSON serialization test for the time being

Issue #542, opened 2022-01-27
[100 SigmaUSD] Wasm and Swift bindings for the arbitrary message signing and verification


greenhat:
### Implementation (DONE)

Add `wallet:signing::sign_message` with the following signature:
```rust
/// Sign arbitrary message under a key representing a statement provable via a sigma-protocol.
/// A statement can be a simple ProveDlog (PK) or a complex sigma conjectives tree
pub fn sign_message(prover: &dyn Prover, sigma_tree: SigmaBoolean, msg: &[u8],) -> Result<Vec<u8>, ProverError> 
```
which should just call the modified `Prover::generate_proof` with removed `ctx_ext` param and returning `ProofBytes` instead of `ProverResult` (should be constructed at the call sites):
```rust
    fn generate_proof(
        &self,
        sigmabool: SigmaBoolean,
        message: &[u8],
        hints_bag: &HintsBag,
    ) -> Result<ProofBytes, ProverError> {
```
Add `Wallet::sign_message` which should just be a wrapper of `wallet::signing::sign_message` (see existing `Wallet::sign_*`).

Add `verifier::verify_signature` with the following signature:
```rust
pub fn verify_signature(sigma_tree: SigmaBoolean, message: &[u8], signature: &[u8]) -> Result<bool, VerifierError>
```
implementation of which should be extracted from `Verifier::verify` (the part after `cprop`).

### Testing (DONE)
Since we're extracting/calling existing code covered with tests a simple smoke test should be enough. Something similar to `verifier::tests::test_prover_verifier_p2pk`.

## Wasm API

Add `wallet::sigma_message_using_p2pk` with the following signature:
```rust
/// Sign arbitrary an message using a P2PK address
pub fn sign_message_using_p2pk(&self, address: Address, message: &[u8]) -> Result<Vec<u8>, JsValue> 
```
Create new `verifier::verify_signature(address: Address, message: &[u8], signature: &[u8]) -> Result<bool, JsValue>`

Create a smoke test in `ergo-lib-wasm/tests` JS tests. 

## Swift API
Add `Wallet.signMessage(withP2PK addr: Address, message: ByteArray) throws -> ByteArray` with all needed underlying functions in `ergo-lib-core-c` and `ergo-lib-c` (see existing `Wallet.signTransaction` for examples).
[...]

PR #553, opened 2022-03-11
Arbitrary message signing and verification (w/o bindings)

nellorocca:
For issue #542, implemented functions for signing messages and verification. 
WASM API and Swift API are not implemented.

### DIFF
--- a/ergo-lib/src/wallet/signing.rs
+++ b/ergo-lib/src/wallet/signing.rs
@@ -274,7 +274,7 @@ pub fn sign_message(
 ) -> Result<Vec<u8>, ProverError> {
     prover
         .generate_proof(sigma_tree, msg, &HintsBag::empty())
-        .map(|p| Vec::from(p))
+        .map(Vec::from)
 }
 
 #[cfg(test)]
--- a/ergo-lib/src/wallet/signing.rs
+++ b/ergo-lib/src/wallet/signing.rs
@@ -470,7 +470,7 @@ mod tests {
 
             // wrong message
             prop_assert_eq!(verify_signature(
-                                            sb.clone(),
+                                            sb,
                                             message.as_slice(),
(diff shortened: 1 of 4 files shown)


ergo-lib/src/wallet/signing.rs
```diff
@@ -260,6 +265,21 @@ pub fn sign_reduced_transaction(
...
+/// Sign arbitrary message under a key representing a statement provable via a sigma-protocol.
+/// A statement can be a simple ProveDlog (PK) or a complex sigma conjectives tree
+pub fn sign_message(
+    prover: &dyn Prover,
+    sigma_tree: SigmaBoolean,
+    msg: &[u8],
+) -> Result<Vec<u8>, ProverError> {
+    let proof = prover.generate_proof(sigma_tree, msg, &HintsBag::empty());
+
+    match proof {
```
greenhat, line 277: I believe the whole `match` expr can be swapped with a single `.map` on `proof`.

ergotree-interpreter/src/sigma_protocol/prover.rs
```diff
@@ -149,7 +149,15 @@ pub trait Prover {
         let ctx_ext = ctx.extension.clone();
         let reduction_result =
             reduce_to_crypto(expr.as_ref(), env, ctx).map_err(ProverError::EvalError)?;
-        self.generate_proof(reduction_result.sigma_prop, message, hints_bag, ctx_ext)
+
+        let proof = self.generate_proof(reduction_result.sigma_prop, message, hints_bag);
+        match proof {
```
greenhat, line 154: Seems like this `match` can be swapped with a `.map` call as well.

greenhat (changes requested):
Looking good! Thank you! Please check my suggestions to use `map`.

greenhat (approved):
Great! Thank you!

PR #554, opened 2022-03-14
add rest feature flag

ross-weir:
- Put `rest` functionality behind feature flag in bindings
- Add `address_from_ergo_tree` functions to bindings

`rest` functionality adds a lot of size to the library which might not always be needed, for example I'm consuming the C library in a go project for some basic functions and I don't need all the rest stuff.

Examples of shared library size differences with MSVC:

with rest -> without rest
release build: 19mb -> 13mb
debug build: 365mb -> 237mb


greenhat:
Good call! Thank you!

bindings/ergo-lib-c/cbindgen.toml
```diff
@@ -140,7 +141,7 @@ extra_bindings = []
 crates = ["ergo-lib-c"]
 all_features = false
```
greenhat, line 142: I've checked all the changes and nothing caught my eye. Although I have not used crate features in C-Swift bindings ever. As a backup plan, maybe `all_features = true` would work.

bindings/ergo-lib-ios/Sources/ErgoLibC/shim.h
```diff
@@ -1,6 +1,11 @@
 #ifndef ERGO_LIB_SHIM_H_
 #define ERGO_LIB_SHIM_H_
 
+// TODO: move this to ios build config
+#ifndef ERGO_REST
```
ross-weir, line 5: Bit of a hack, this should somehow be defined in the ios build config

Having no ios experience and no hardware I'm not able to figure out how to do this

greenhat: I have no idea for a better place for this `define`. Let's leave it here for now.

greenhat (approved):
Great job! I appreciate your efforts! Thank you!

Issue #552, opened 2022-03-08
ExtPubKey derivation differs from Ledger and bip32 library


greenhat:
from our discord:

>Getting the public key and the chain code from the ledger device I was not able to generates the same ergo addresses than the ones produced restoring the mnemonic of the ledger in a wallet.
I've used the bip32 library to workaround the issue.
Here the code change fixing the issue for me: https://github.com/ThierryM1212/SAFEW/commit/02e37ec5946391494ab270906796382226eb8d5e#diff-d7b533269ae809e99d492632b6ddc2ceff09e33596a7c5e4775f34b1e11e5ea5L11-R22

>The addresses generated from wasm library are the same than the ones from bip32 lib, if the ".derive(index)" is removed.
In the wasm code, I tried to add additionals "derive(path)" or "child(index)", but the generated addresses were not fixed.

>I'm not really able to create a test case... I would need a way to export the pubkey and chain_code from an ExtPubKey.
(Yes... first complains about possible bugs without testcase and then ask for new features... sorry 😉 )

PR #556, opened 2022-03-25
Add pubkey & chain code methods

ross-weir:
Add some of the missing methods from #552 - haven't looked into the bug yet though


greenhat (approved):
Thank you!

Issue #545, opened 2022-02-04
`/nipopow/proof` REST API endpoint support


greenhat:
Implement `ergo_rest::api::node::get_nipopow_proof_by_header_id` using `/nipopow/proof/{m}/{k}/{headerId}` REST API endpoint.

Add it to the Wasm and C/Swift API. See existing `node::get_info` code for examples.

kettlebell:
Will be done as part of #544.

PR #558, opened 2022-04-01
Node discovery via node's REST API

### LINKED ISSUE #544: [350 SigmaUSD] Node discovery via node's REST API
### Implementation details

Take a list of nodes (seeds) and ask them for their known peers via `/peers/all` REST API endpoint. Repeat the same operation (ask for their known peers) with received nodes skipping the ones we've already asked, keeping a limit of simultaneous requests. Test if the REST API is available on the fetched nodes via `/info` endpoint on port 9053. Filter out nodes that do not have REST API(connection refused, timeout, etc.) Stop when there are no nodes left to ask or a given timeout is reached.
Return the list of discovered nodes without the seeds.

### Rust API
```rust
async fn peer_discovery(seeds: NonEmptyVec<Url>, max_parallel_requests: BoundedU16<1, u16:MAX>, 
    timeout: Duration) -> Result<Vec<Url>, PeerDiscoveryError>
```
Introduce `type NonEmptyVec<T> = BoundedVec<T, 1, usize::MAX>;` in `bounded-vec` crate.
Testing should be done by running a peer discovery on the public nodes list(seeds) shipped with the node - https://github.com/ergoplatform/ergo/blob/f38f3a6d0ab8849871adc54010846c5afb101ff4/src/main/resources/mainnet.conf#L49-L64 using port 9053 (default for REST API).

### Wasm API
```rust
async fn peer_discovery(seeds: Box<[web_sys::Url]>, max_parallel_reqs: u16,
    timeout_sec: u32) -> Result<Box<[web_sys::Url]>, JsValue>
```

### Swift API
```swift
func peerDiscovery(seeds: [URL], maxParallelReqs:UInt16, timeoutSec: UInt32, 
    closure: @escaping (Result<[URL], Error>) -> Void) throws -> RequestHandle
```
The underlying C FFI layer should use string for `URL` representation, passing the array of strings as a pointer and a count.

`/nipopow/proof` REST API endpoint support

Implement `ergo_rest::api::node::get_nipopow_proof_by_header_id` using `/nipopow/proof/{m}/{k}/{headerId}` REST API endpoint.

Add it to the Wasm and C/Swift API. See existing `node::get_info` code for examples.

kettlebell:
Closes #544 and and #545


bindings/ergo-lib-ios/Tests/ErgoLibTests/RestNodeApiTests.swift
```diff
@@ -1,49 +1,103 @@
...
+                        break
+                    case .failure(let error):
+                        XCTFail(error.localizedDescription)
+                }
+                expectation.fulfill()
+            })
+        waitForExpectations(timeout: 60, handler: nil)
+    }
+    
+    func testPeerDiscoveryAsync() async throws {
```
kettlebell, line 85: The async tests mistakenly result in 'pass' when using `Task` in the unit tests. The fix is to make the entire function async and avoid making calls to `Task`.

bindings/ergo-lib-wasm/src/rest/api.rs
```diff
@@ -14,3 +15,72 @@ pub async fn get_info(node: NodeConf) -> Result<JsValue, JsValue> {
...
+        self.0.len()
+    }
+
+    /// Returns the element of the collection with a given index
+    pub fn get(&self, index: usize) -> web_sys::Url {
+        self.0[index].clone()
+    }
+}
+
+/// Stub that can be removed once `bounded-vec` is updated with this type
```
kettlebell, line 58: Got this stub in various places. Perhaps we can bump the version for `bounded-vec`?

bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
```diff
@@ -0,0 +1,43 @@
...
+        "http://51.75.147.1:9020",
+        "http://221.165.214.185:9030",
+        "http://51.81.185.231:9031",
+        "http://217.182.197.196:9030",
+        "http://62.171.190.193:9030",
+        "http://173.212.220.9:9030",
+        "http://176.9.65.58:9130",
+        "http://213.152.106.56:9030",
+    ].map(x => new URL(x));
+    let res = await ergo_wasm.peer_discovery(seeds, 10, 3);
```
kettlebell, line 41: Surprisingly the `max_parallel_requests` arg makes a big difference in the runtime here. My understanding is that JS is strictly single-threaded (aside from things like `web-workers`) but I guess async still gives nice gains when waiting for IO to complete.

kettlebell:
Everything works and I am happy to report that there's actually about 13-14 active nodes with an open REST API! As an aside @MrStahlfelge has a nice page that reports the peers https://api.tokenjay.app/peers/list, and my implementation is consistent with it. Note that setting the timeout to 1 second results in some peers not being discovered. 2 seconds seems to be enough.

ergo-rest/src/api/node.rs
```diff
@@ -46,18 +47,437 @@ pub async fn get_info(node: NodeConf) -> Result<NodeInfo, NodeError> {
...
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+/// Given a list of seed nodes, recursively determine a Vec of all known peer nodes. Note that there
+/// are `println` calls below that are commented out. Uncommenting them will show the requests that
+/// are made, especially when made in parallel.
+pub async fn peer_discovery(
+    seeds: NonEmptyVec<Url>,
+    max_parallel_requests: BoundedU16<1, { u16::MAX }>,
+    timeout: Duration,
```
greenhat, line 66: Initially, I was planning for `timeout` to be an "overall" timeout for the whole peer discovery run. In case there are way too many peers to discover them all in a reasonable amount of time. 
We might get away without it now since the number of active nodes is quite low but I'd rather have it implemented now.
What do you think?

greenhat (changes requested):
Looking great! Awesome job! Please see a few minor comments.

bindings/ergo-lib-wasm/src/rest/api.rs
greenhat: Yes, please do.

bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
greenhat: I think so as well. I wonder if we should advise any upper limit in the comments.

bindings/ergo-lib-wasm/src/rest/api.rs
greenhat: Oh, I think you meant that I should release a new version of bounded-vec, right? :) I'll do it ASAP.

bindings/ergo-lib-wasm/package.json
```diff
@@ -6,12 +6,11 @@
...
-        "integration-test": "wasm-pack build --target nodejs --features rest && mocha --require @babel/register --require @babel/polyfill 'tests/integration_test*.js'",
         "test-browser": "rm -rf ./pkg && npm run build-browser && mv ./pkg-browser ./pkg && ./node_modules/.bin/karma start",
         "doc": "jsdoc -c jsdoc.json pkg/ergo_wallet_wasm.js README.md -d docs",
         "build-nodejs": "rm -rf ./pkg-nodejs && wasm-pack build --target nodejs --out-dir pkg-nodejs && cd pkg-nodejs && node ../scripts/publish_helper -nodejs",
         "build-nodejs-alpha": "rm -rf ./pkg-nodejs && wasm-pack build --target nodejs --out-dir pkg-nodejs && cd pkg-nodejs && node ../scripts/publish_helper -nodejs && npm version minor && node ../scripts/set_alpha_version -nodejs $(git rev-parse --short HEAD)",
-        "build-browser": "rm -rf ./pkg-browser && wasm-pack build --target browser --out-dir pkg-browser && cd pkg-browser && node ../scripts/publish_helper -browser",
+        "build-browser": "rm -rf ./pkg-browser && wasm-pack build --target browser --out-dir pkg-browser --features rest && cd pkg-browser && node ../scripts/publish_helper -browser",
```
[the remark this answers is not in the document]
kettlebell: Yes of course, will do.

ergo-rest/src/api/node.rs
kettlebell: Yeah that's a great idea. Perhaps the timeout can be a 'best-effort' one: once timeout is reached, return the peers discovered up to that point in time. Sounds ok?

greenhat: Exactly!

bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
kettlebell: I set some pretty high levels (30+), and it never complained. Since we live in a world where we have hundreds of tabs in our browsers, I'm not sure what's a reasonable limit!

greenhat: Yeah, I have the same feeling.

ergo-rest/src/api/node.rs
kettlebell: I've implemented it, but it requires us to vendor another crate. To implement timeout we need an async way to sleep. Tokio already has this built-in but for WASM there isn't a standard to rely on.

There are 2 possible crates that do this: `futures-timer` and `wasm-timer`. Both of them in current form break the `webpack` build process. I couldn't figure out how to fix it with `futures-timer`, but tracked it down with `wasm-timer`. Basically, there's an out of date dependency that is at fault. Now this crate hasn't been updated since 2020, hence my vendoring. Hope this is ok.

ergo-rest/src/api/node.rs
```diff
@@ -46,18 +47,437 @@ pub async fn get_info(node: NodeConf) -> Result<NodeInfo, NodeError> {
...
+}
+
+#[cfg(target_arch = "wasm32")]
+/// Given a list of seed nodes, recursively determine a Vec of all known peer nodes.
+///
+/// This version is essentially the same as the tokio-based, non-WASM version above except that
+/// the `spawn_local` function can only spawn tasks for futures that return `()`. This means that
+/// we can't do the usual error handling. We can handle some errors by sending out occurrences
+/// through a spawned channel.
+pub async fn peer_discovery(
```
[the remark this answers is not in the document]
kettlebell: I've implemented this, restructured the code and added docs.

ergo-rest/src/api/peer_discovery_internals.rs
```diff
@@ -0,0 +1,410 @@
...
+    url_stream: impl futures::Stream<Item = Url> + Send + 'static,
+    max_parallel_requests: BoundedU16<1, { u16::MAX }>,
+    request_timeout_duration: Duration,
+) {
+    use futures::StreamExt;
+
+    // Note that `tokio` - the de facto standard async runtime - is not supported on WASM. We need
+    // to spawn tasks for HTTP requests, and for WASM we rely on the `wasm_bindgen_futures` crate.
+    #[cfg(not(target_arch = "wasm32"))]
+    let spawn_fn = tokio::spawn;
```
kettlebell, line 239: Wasn't able to abstract out the spawn functions with traits (just couldn't type check it). Used conditional compilation instead. It's ugly but it works.

ergo-rest/src/api/peer_discovery_internals.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// to the node's /info endpoint responds with code 200 OK.
+    AddActiveNode(Url),
+    /// Indicates that the ergo node at the given URL is inactive. This means that a GET request
+    /// to the node's /info endpoint does not respond with code 200 OK.
+    AddInactiveNode(Url),
+    /// A list of peers of an active ergo node, returned from a GET on the /peers/all endpoint.
+    CheckPeers(Vec<PeerInfo>),
+}
+
+/// This trait abstracts over the `send` method of channel senders, assuming no failure.
```
kettlebell, line 323: There are 2 types of channels to abstract over: `tokio::sync::mpsc::...` and `futures::channel::mpsc...`, hence these traits.

wasm-timer/README.md
```diff
@@ -0,0 +1,7 @@
+# Wasm-timer
+
+**This is a simple fork of the original repo at https://github.com/tomaka/wasm-timer**. The crate has
```
kettlebell, line 3: Here I explain the need for the vendoring.

kettlebell:
I've made the requested changes!

greenhat (changes requested):
Looking great! I appreciate the explanations! Thank you! Only a minor folder structure comment.

greenhat (approved):
Thank you!

PR #562, opened 2022-04-13
Serialization for `NipopowProof`

### LINKED ISSUE #490: [100 SigmaUSD] Add NipopowProof type
Scala reference implementation -
https://github.com/ergoplatform/ergo/blob/95139ec990fc91f8f4c70a61468edf8834143329/src/main/scala/org/ergoplatform/modifiers/history/popow/NipopowProof.scala#L25-L30
(ignore NipopowAlgos and skip NipopowProof methods)

In ergo-nipopow crate implement binary (implement ScorexSerializable trait) and JSON
serialization.

kettlebell:
Closes #490

### DIFF
--- a/ergo-rest/src/api/node.rs
+++ b/ergo-rest/src/api/node.rs
@@ -92,6 +92,6 @@ mod tests {
             timeout: Some(Duration::from_secs(5)),
         };
         let res = runtime_inner.block_on(async { get_info(node_conf).await.unwrap() });
-        assert_eq!(res.name, "ergo-mainnet-4.0.16.1");
+        assert_ne!(res.name, "");
     }
 }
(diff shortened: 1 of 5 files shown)


ergo-rest/src/api/node.rs
```diff
@@ -92,6 +92,6 @@ mod tests {
             timeout: Some(Duration::from_secs(5)),
         };
         let res = runtime_inner.block_on(async { get_info(node_conf).await.unwrap() });
-        assert_eq!(res.name, "ergo-mainnet-4.0.16.1");
```
kettlebell, line 95: The node now replies with "ew" as its name. We can't anticipate any future changes here, so check that it's not empty.

greenhat: Thanks! My bad, for making the test so fragile. I was very happy it worked so I forgot to make a more sensible check. :)

greenhat (approved):
Looking great! Thank you!

ergo-rest/src/api/node.rs
kettlebell: I know the feeling!

PR #565, opened 2022-04-21
Ban unwrap in `ergo-lib-wasm`

### LINKED ISSUE #532: Ban `unwrap` in ergo-lib-wasm
Add the following to `lib.rs`:
```rust
#![deny(clippy::unwrap_used)]
#![deny(clippy::expect_used)]
#![deny(clippy::todo)]
```
Most of the errors can be eliminated with error handling. The cases where `unwrap` is necessary should be enabled with `allow` attribute with a comment explaining the reason.

kettlebell:
Close #532

### DIFF
--- a/bindings/ergo-lib-wasm/src/json.rs
+++ b/bindings/ergo-lib-wasm/src/json.rs
@@ -37,14 +37,12 @@ pub(crate) struct TransactionJsonEip12 {
 
 impl From<Transaction> for TransactionJsonEip12 {
     fn from(t: Transaction) -> Self {
-        // Following unwraps are fine since we're converting from BoundedVec to Vec.
-        #[allow(clippy::unwrap_used)]
         TransactionJsonEip12 {
             tx_id: t.id(),
-            inputs: t.inputs.try_into().unwrap(),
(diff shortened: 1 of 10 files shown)


bindings/ergo-lib-wasm/src/json.rs
```diff
@@ -37,6 +37,8 @@ pub(crate) struct TransactionJsonEip12 {
 
 impl From<Transaction> for TransactionJsonEip12 {
     fn from(t: Transaction) -> Self {
+        // Following unwraps are fine since we're converting from BoundedVec to Vec.
```
greenhat, line 40: I'm not sure why try_into() is used here. It's an infallible op. There is as_vec() and From implemented for getting the Vec out of BoundedVec.

greenhat (changes requested):
Looking great! Thank you! Just a minor comment.

bindings/ergo-lib-wasm/src/json.rs
kettlebell: Thanks, making the change now!

greenhat (approved):

PR #568, opened 2022-04-27
update docs for building on ios

aostiles:
The cbindgen config seems to default to generating headers for the "rest" feature (line 144). The swift bindings also seem to assume the built `libergo` has the rest symbols defined. I found it a little strange because `ergo-lib-c` doesn't build the rest feature by default. Maybe there's some context I'm missing.

I also suggested skipping the REST API tests by default. I was just trying to get the library to build and didn't have an ergo node running.

### DIFF
--- a/bindings/ergo-lib-ios/README.md
+++ b/bindings/ergo-lib-ios/README.md
@@ -5,7 +5,7 @@
 
 First build `ergo-lib-c`
 ```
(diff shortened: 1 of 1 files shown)


greenhat:
Welcome! Thank you for your contribution!

@ross-weir  Could you please take a look?

PR #571, opened 2022-04-30
Convert vendored dependencies of `ergo-rest` into sub-modules 

### LINKED ISSUE #570: Fix vendored-deps crates so ergo-rest can be published
On publishing dependencies that are specified by `path` are taken by `version` with `path` ignored.
I see two options:
1) Embed the vendored crates source files as modules of ergo-rest crate;
2) Publish vendored crates under an altered name;

I'd try 1) first and use 2) only as last resort.

kettlebell:
Close #570


.github/workflows/ci.yml
```diff
@@ -142,11 +142,6 @@ jobs:
           curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
           sudo apt-get -y install nodejs
       
-      - name: run `reqwest-wrap` WASM tests
```
kettlebell, line 145: I had these tests here originally to test the timeout PR that was applied on top of `reqwest`. After moving it under `ergo-rest`, the WASM tests for it no longer compiles, due to a really weird error in the dependency hierarchy (for reference it's exactly this error message: https://github.com/alexcrichton/wait-timeout/issues/25).

I've opted to delete it for now.

ergo-rest/src/wasm_timer/timer/ext.rs
```diff
@@ -23,30 +23,6 @@ pub trait TryFutureExt: TryFuture + Sized {
     /// If the future completes before `dur` elapses then the future will
     /// resolve with that item. Otherwise the future will resolve to an error
     /// once `dur` has elapsed.
-    ///
```
kettlebell, line 26: I deleted this doctest since otherwise we'd leak some `wasm-timer` internals through `ergo-rest`.

greenhat (changes requested):
Looking great! I appreciate your description of submodules in the readme file. Thank you!
One minor request - to exclude vendored sub-modules from test coverage tracking. Could you please add them to the existing `--exclude-files` option in ci.yml?

kettlebell:
Made the change.

greenhat (approved):
Thank you!

PR #572, opened 2022-05-03
Move `Header` type from `ergotree-ir` to `ergo-chain-types`

### LINKED ISSUE #488: [100 SigmaUSD] Move Header type from ergotree-ir to ergo-chain-types crate
Along with JSON serialization and `Arbitrary` implementation (behind `arbitrary` feature).

kettlebell:
Close #488


ergotree-ir/src/chain/header.rs
```diff
@@ -1,502 +1,2 @@
...
-            header.autolykos_solution.pow_distance,
-            Some(BigInt::from_str(
-                "1234000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-            )
-            .unwrap())
-        );
-    }
-}
+//! Block header and AutolykosSolution
+pub use ergo_chain_types::{AutolykosSolution, Header};
```
greenhat, line 2: I'm not sure it's a good idea. Somewhere a few crates up in the dep tree when dev will use Header they will be presented with a choice of two (original and re-exported).

EDIT: The same goes to all moved and re-exported types in ergotree-ir.
EDIT2: I understand that re-export allows us to avoid breaking user's code. I think we're can brake it in this case.

ergotree-ir/src/chain/preheader.rs
```diff
@@ -26,6 +22,20 @@ pub struct PreHeader {
     pub votes: Votes,
 }
 
+impl From<Header> for PreHeader {
```
greenhat, line 25: Any chance `PreHeader` can be moved as well? 

EDIT: I don't see a reason for it to stay here.

greenhat (changes requested):
Looking great! Thank you! Only re-export bothers me.

ergotree-ir/src/chain/preheader.rs
kettlebell: Done.

PR #574, opened 2022-05-04
Refactor `TransactionContext`

### LINKED ISSUE #497: `TransactionContext::get_boxes_to_spend` is dangerous
Since the order of the tx inputs could be different we should eliminate `TransactionContext::get_boxes_to_spend()` and introduce `TransactionContext::get_input_box(box_id: BoxId) -> Option<ErgoBox>`.

Besides that, we should move `TransactionContext` to a submodule `tx_context` but leave the export `pub use tx_context::TransactionContext` in the `signing` module to not break the existing imports. So that no function in `signing` would have the access to private `TransactionContext::boxes_to_spend` property.

While we are at it let's change `TransactionContext::new(...)` to accept `Vec` for inputs and do the check here (see call sites from Wasm layer for the unwanted boilerplate).

kettlebell:
Close #497

### DIFF
--- a/bindings/ergo-lib-c-core/src/reduced.rs
+++ b/bindings/ergo-lib-c-core/src/reduced.rs
@@ -2,7 +2,7 @@
 //! is augmented with ReducedInput which contains a script reduction result.
 
 use ergo_lib::{
-    chain::transaction::{reduced::reduce_tx, TxIoVec},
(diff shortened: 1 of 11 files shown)


greenhat (approved):
Looking great! Thank you!

PR #584, opened 2022-06-10
Fix `peer_discovery` on Chrome

### LINKED ISSUE #581: [500 SigmaUSD] `peer_discovery` REST API has problems on Chrome
The current implementation of `peer_discovery` doesn't quite work on Chrome (it works perfectly on Firefox and tokio). Basically it cannot discover anywhere near the amount of existing active nodes (it mostly finds 0 or 1). I tried my best to find the underlying cause, but no dice. Although the function returns without error, there is also a really bad bug that results: after calling `peer_discovery` any subsequent HTTP requests fail, with `reqwest` reporting that the request times out.

## Mitigation
I tweaked the value of many parameters and found a mitigation for this. My educated guess from these experiments is that Chrome has an issue where it can't handle too many concurrent promises. So referring to my schematic from the code, `peer_discovery` works as follows:

 - Task 1 is responsible for tracking which nodes are active/inactive and making sure that any
   given ergo node is queried exactly once.
 - Task 2's job is to wait for a URL from task 1, make the actual HTTP requests to that URL, and
   to report the result back to task 1.
```text
                             <ergo node URL>
              __________________________________________________
             |                                                  |
             |                                                  v
 /----------------------\                   /----------------------\
 | 1. Track node status |                   | 2. HTTP request task |
 \----------------------/                   \----------------------/
             ^                                                  |
             |__________________________________________________|
               <active node| non-active node| list of peers>   
```

The solution for Chrome is to restrict the number of URLs sent to task 2. More specifically we must make the following changes:
 - The global timeout interval is key. For chrome this needs to be **no more than 1.5 seconds**. Anything longer and the bug described above appears.
 - The timeout isn't strict anymore. When it triggers we wait for any in-flight URLs to be checked, then we return.
 - The queue that task 2 uses to buffer incoming URLs is set to length of 50.
 - Surprisingly, the timeout interval for each individual HTTP request to each peer node needs to be about 6 seconds (this even affects Firefox). Lower values results in missing some active nodes. This problem doesn't occur with the `tokio` version: things work fine setting the interval to 2 seconds.

Now what this means is that Chrome can't discover all the available nodes (it can get about 13). The following table gives rough results by timeout length when using Firefox:

|Global timeout(sec)| # peers reached| # active peers discovered| 
|--|---|--|
|[1.0, 4.4] | ~260 | ~13|
| [4.5, ..) | ~560| ~16|

The results are like a step function. For example any interval between 1 to 4.4 seconds gives us 13 active nodes.

kettlebell:
Close #581

Summary of changes:
 - Implemented a chrome-specific version of `peer_discovery`, as well as an incremental version that remembers the results from previous calls. A drawback of the incremental version is that no matter how small the sub-task, there's always an 80 second wait time at the end to make sure the preflight requests are relinquished by Chrome. I could give an option to return early, but the end user needs to be careful to not eat up all the requests.
 - Made an improvement to the `reqwest` timeouts for WASM, where if the HTTP request is successful, the associated timeout is cancelled.
 - The JS browser tests now use `peer_discovery`, so no more hard-coded node addresses. The tests are quite lengthy though; the incremental `peer_discovery` test is about 8 minutes.


bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
```diff
@@ -10,8 +10,32 @@ beforeEach(async () => {
...
+    scan = await ergo_wasm.incremental_peer_discovery_chrome(scan, 150, 90);
+    let scan_1_len = scan.active_peers().len();
+    scan = await ergo_wasm.incremental_peer_discovery_chrome(scan, 150, 400);
+    let scan_2_len = scan.active_peers().len();
+    assert(scan_1_len < scan_2_len, "Should have found more peers after second scan!");
+});
+
 it('node REST API: get_nipopow_proof_by_header_id endpoint', async () => {
-    let node_conf = new ergo_wasm.NodeConf("213.239.193.208:9053");
+    let peers = await ergo_wasm.peer_discovery_chrome(get_ergo_node_seeds(), 150, 150);
```
greenhat, line 36: I'm not sure we should use live peer discovery in tests. Only in tests of the peer discovery itself (two tests above). In all other tests, a hardcoded node list should suffice.

greenhat (changes requested):
Looking awesome! Great job! Please check my comments. 
I've underestimated this task, so I'm increasing the bounty.

bindings/ergo-lib-wasm/src/rest/api.rs
```diff
@@ -125,11 +125,15 @@ impl PeerUrls {
...
-///  - `max_parallel_requests` represents the maximum number of HTTP requests that can be made in
-///    parallel
+///  - `max_parallel_tasks` represents the maximum number of tasks to spawn for ergo node HTTP
+///    requests. Note that the actual number of parallel HTTP requests may well be higher than this
+///    number.
 ///  - `timeout` represents the amount of time that is spent search for peers. Once the timeout
 ///    value is reached, return with the vec of active peers that have been discovered up to that
 ///    point in time.
+///
+/// IMPORTANT: do not call this function on Chromium, as it will likely mess with the browser's
```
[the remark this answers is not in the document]
kettlebell: Thanks for the suggestion, I've implemented it plus copied over extra explanations from the `ergo-rest`.

bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
greenhat: Yeah, that's not good. However, running peer discovery in these tests looks like a huge overkill. Plus, I'm not sure how github views if we shoot a lot of network requests in GA. It might be a violation of their TOS.
What do you think of running a peer discovery once per test session and making discovered peers list accessible in relevant tests?

kettlebell: That's a great point concerning the TOS. I've implemented what you've suggested and it's surfaced an issue specific to github CI where many active peers aren't getting picked up compared with running it locally. I think it's due to the timeout interval of each individual request (it's currently set to 6 seconds). You can see I've made a bunch of noisy commits where I'm experimenting, so sorry about that. I hope to resolve this soon!

kettlebell: I just did a rough test, and my guess doesn't seem right. With the incremental test, the table below shows the number of active nodes picked up by Github and locally, for each incremental round:
|   | Github Actions| Locally|
|---|---|--|
|increment 1| 0| 6|
| increment 2| 2 | 21|

So there's something going on with Github actions. Is it ok for me to clean things up, set hardcoded IPs again for the SPV test and leave it here? Then I can finally focus on the oracle-core tasks : )

greenhat: Sure. No need to dig further with CI.

kettlebell: Thanks. Please note the current error in CI, where the 2nd incremental round found no new peers. It's unfortunate, as it never happens locally.

bindings/ergo-lib-wasm/src/rest/api.rs
```diff
@@ -165,16 +180,32 @@ pub async fn peer_discovery(
...
+/// and running our first implementation of `peer_discovery` in the `non_chrome` submodule brings
+/// Chrome to a halt, preventing it from making any further requests until enough of the preflights
+/// have been marked as failed. This is quite unfortunate as Firefox and Safari almost immediately
+/// drops such preflight requests. See [`this issue`] for more discussion and example runs.
+//
+/// So this implementation is largely the same as in `non_chrome` except that we throttle the number
+/// of parallel requests made. The majority of ergo nodes do not respond to REST requests, so there
+/// will be a large number of preflight requests 'taking up space' and dramatically reducing
+/// throughput.
 pub async fn peer_discovery_chrome(
```
greenhat, line 209: Should it be private?

greenhat (changes requested):
Great! Thank you! Just a tiny API visibility question.

bindings/ergo-lib-wasm/src/rest/api.rs
kettlebell: Of course, thanks!

bindings/ergo-lib-wasm/tests_browser/test_rest_api.js
kettlebell: After latest commit, error is in the same place. I made issue #586 for this. As a temporary measure, we can allow the 2nd round to find no new peers, just to silence the CI errors that will crop up from everyone's commits. What do you think?

kettlebell: Actually, it looks like the problem comes about from persisting `peer_discovery` results between tests. Reverted this, and `peer_discovery` starts finding more active nodes. Looks like this is fixable. Never a boring day with web dev...

kettlebell: It's still nowhere near the performance of running it locally though:
image

So a single round of `peer_discovery` with search time 60 seconds yields 3 active nodes. For the incremental test, the first round is 10 second search time, and the second round is 400 seconds search time. The second round only finds one extra node, which isn't good at all.

But I think everything's ready for merging though!

greenhat: Awesome!

greenhat (approved):
Great job! Thank you!

PR #591, opened 2022-07-10
Fix for #590 accidental token burn in change box

### LINKED ISSUE #590: Unwanted token burn incident in Nautilus wallet
from discord https://discord.com/channels/668903786361651200/729692906209673257/995038064277737502

> A user reported on nautilus channel that he got a big amount of tokens burned while doing a regular transaction.

> After some investigation I found out that it could be some bug on sigma-rust. The wallet selected three unspent boxes as input with 275 distinct tokens in total and assembled the transaction using sigma-rust, for some reason the tokens that were supposed to included into the change box were not included and ended up getting burned.

> I managed to simulate the transaction and it looks like the issue is related with the limit of 255 tokens per box, if I manually remove some tokens from the inputs, the change tokens gets included normally into the change box. 

> Tested versions: v0.6.1 and v0.7.0

> Please, let me know if you need some additional clarification.

> Here is a test case simulating this issue: https://gist.github.com/capt-nemo429/16f59d6bd6f3e86209b2060e0cb685c9

> Transaction: https://explorer.ergoplatform.com/en/transactions/6f2e8307f74cec87566ae59e9b743c799e48c6645633f30d3c9df827bbb0c573 

> Here is the transaction generated by the code I've shared above: 

```

{
  "inputs": [
    {
      "boxId": "490148afdc36f5459bbfd84922a446abea9a1077e031822f377b0ff3a6e467e3",
      "extension": {}
    },
    {
      "boxId": "f0b6ff61c41692ddbb4a57acd8ff5ca7f113bbd89278c28535f8347b96279967",
      "extension": {}
    },
    {
      "boxId": "a9ca7b041922d1051c01196610b64eb61f17dbe1951a4e638620da7fba0a2855",
      "extension": {}
    }
  ],
  "dataInputs": [],
  "outputs": [
    {
      "value": 1000000,
      "ergoTree": "0008cd037fae590b9610a6d0f8634184b7dc4a76de8610a398492106de74b175e93a5412",
      "assets": [
        {
          "tokenId": "7d1568dce627141e6c9945328ef2bae5b7cc5c4f67fd873e288b808e68447db3",
          "amount": 1
        },
        {
          "tokenId": "45a773bb751f4a1cb652624026a054b1fd7e526896d9b566938398d9bbdf2716",
          "amount": 1
        },
        {
          "tokenId": "7d441ce250ffd0d70511899e5e9090f20a0443260983f76ef89a32cbb3eabda4",
          "amount": 1
        },
        {
          "tokenId": "e21d27d9bbfec0a385e33c8987f62d14955ade38b3030d97681868b7492c76ad",
          "amount": 1
        },
        {
          "tokenId": "1a77f5e1f185318c16c13402b4df133d439ab0e85382eab0b5598c7a8aa6a6c9",
          "amount": 1
        }
      ],
      "additionalRegisters": {},
      "creationHeight": 788658
    },
    {
      "value": 7765648,
      "ergoTree": "0008cd02f469fd785dbbc9112514655250c414d0c122d7df5eabc6966985b45b34542c02",
      "assets": [],
      "additionalRegisters": {},
      "creationHeight": 788658
    },
    {
      "value": 1100000,
      "ergoTree": "1005040004000e36100204a00b08cd0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798ea02d192a39a8cc7a701730073011001020402d19683030193a38cc7b2a57300000193c2b2a57301007473027303830108cdeeac93b1a57304",
      "assets": [],
      "additionalRegisters": {},
      "creationHeight": 788658
    }
  ]
}
```

greenhat:
Close #590 

Done in this PR:

- reproduce submitted in #590 code as JS test;
- reproduce the issue in Rust property-based test with various tokens count;
- fix the issue in the box selector and implement splitting the change into multiple boxes (see `make_change_boxes()`);
- add a guard that checks `selected inputs = target + change` for ERG and tokens at the end of the box selection process (see `check_input_preservation()`);
- refine `BoxSelectorError` for comprehensible error messages(remove generic error conversion in favor of tailored messages for each error case);

### DIFF
--- a/bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
+++ b/bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
@@ -44,14 +44,16 @@ final class TokenIdTests: XCTestCase {
         let amount = Int64(12345678)
         let tokenAmount = try TokenAmount(fromInt64: amount)
         let token = Token(tokenId: tokenId, tokenAmount: tokenAmount)
+
+        let maxTokensCount = 100
         
-        for _ in 1...255 {
+        for _ in 1...maxTokensCount {
             try tokens.add(token: token)
         }
-        XCTAssertEqual(tokens.len(), UInt(255))
-        XCTAssertNotNil(tokens.get(index: 254))
+        XCTAssertEqual(tokens.len(), UInt(maxTokensCount))
+        XCTAssertNotNil(tokens.get(index: maxTokensCount - 1))
         
-        // Add 256th token, expecting error
+        // Add max + 1 token, expecting error
         XCTAssertThrowsError(try tokens.add(token: token))
     }
 }
--- a/bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
+++ b/bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
@@ -45,7 +45,7 @@ final class TokenIdTests: XCTestCase {
         let tokenAmount = try TokenAmount(fromInt64: amount)
         let token = Token(tokenId: tokenId, tokenAmount: tokenAmount)
 
-        let maxTokensCount = 100
+        let maxTokensCount = UInt(100)
         
         for _ in 1...maxTokensCount {
             try tokens.add(token: token)
--- a/ergo-lib/src/wallet/box_selector/simple.rs
+++ b/ergo-lib/src/wallet/box_selector/simple.rs
@@ -170,11 +170,14 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
 mod tests {
     use std::convert::TryFrom;
 
-    use ergotree_ir::chain::ergo_box::box_value::checked_sum;
     use ergotree_ir::chain::ergo_box::ErgoBox;
+    use ergotree_ir::chain::{ergo_box::box_value::checked_sum, token::arbitrary::ArbTokenIdParam};
     use proptest::{collection::vec, prelude::*};
 
-    use crate::wallet::box_selector::sum_value;
+    use crate::wallet::box_selector::{
+        arbitrary::{ArbErgoBoxAssetsDataParam, ArbTokensParam},
+        sum_value,
+    };
 
     use super::*;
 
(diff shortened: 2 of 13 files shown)


ergo-lib/src/wallet/box_selector/simple.rs
```diff
@@ -152,29 +166,144 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
...
+                BoxTokens::from_vec(change_tokens.into_iter().map(Token::from).collect()).unwrap(),
+            ),
+        }])
+    } else {
+        let mut change_boxes = vec![];
+        let mut change_tokens_left: Vec<Token> =
+            change_tokens.into_iter().map(Token::from).collect();
+        let mut change_value_left = change_value;
+        while !change_tokens_left.is_empty() {
+            if change_tokens_left.len() < ErgoBox::MAX_TOKENS_COUNT {
```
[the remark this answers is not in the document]
kettlebell: Nice spot.

kettlebell (changes requested):
Looks great to me, aside from the issue @SethDusek found!

ergo-lib/src/wallet/box_selector/simple.rs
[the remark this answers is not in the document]
greenhat: Good catch! Thank you! I guess I subconsciously tried to avoid the edge case (255 tokens). Fixed here and above(whole box, no split) and added an `ErgoBox` test with 255 tokens.

ergo-lib/src/wallet/box_selector/simple.rs
```diff
@@ -152,29 +166,144 @@ impl<T: ErgoBoxAssets> BoxSelector<T> for SimpleBoxSelector {
...
+                    value: change_value_left,
+                    tokens: Some(BoxTokens::from_vec(change_tokens_left).unwrap()),
+                };
+                change_boxes.push(change_box);
+                break;
+            } else {
+                #[allow(clippy::unwrap_used)] // safe for the box value upper bound
+                // doubled due to larger box size to accomodate so many tokens
+                let value = BoxValue::SAFE_USER_MIN.checked_mul_u32(2).unwrap();
+                let tokens_to_drain = ErgoBox::MAX_TOKENS_COUNT / 2;
```
SethDusek, line 269: Small nitpick, but I think a more efficient way to distribute boxes would be to drain change_tokens.len() % (ErgoBox::MAX_TOKENS_COUNT + 1) each time. Right now if a user has say, 510 Tokens, the distribution will be done like so if I've traced the code correctly:

First Box has 127 tokens (MAX_TOKENS_COUNT / 2), 383 still remaining
Second Box has 127 tokens, 256 remaining
Third Box will have 127 tokens, 129 remaining
Fourth one will have 129

Whereas this could be done with two boxes, each having 255 tokens.

SethDusek (approved):

greenhat:
@TimL @SethDusek It turned out the 255 token limit is wrong. We have a max box size limit of 4k and since token id is 32 bytes alone there are only about ~110 tokens fit in the box. I missed it by overly relying on this comment in box serialization - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/6392cbcbe14988e6465bcd7d0ea43242abb109b5/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeTransaction.scala#L123-L126
which is contradicted by this check in the transaction verification on the node - https://github.com/ergoplatform/ergo/blob/a4d1e62fb8f888a33db3fc5881c2090ec3dd0ebb/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala#L168

So I set the token count limit to 100, mirroring the node wallet's approach in 
https://github.com/ergoplatform/ergo/blob/a4d1e62fb8f888a33db3fc5881c2090ec3dd0ebb/ergo-wallet/src/main/scala/org/ergoplatform/wallet/Constants.scala#L43-L46

bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
```diff
@@ -44,14 +44,16 @@ final class TokenIdTests: XCTestCase {
...
+
+        let maxTokensCount = UInt(100)
         
-        for _ in 1...255 {
+        for _ in 1...maxTokensCount {
             try tokens.add(token: token)
         }
-        XCTAssertEqual(tokens.len(), UInt(255))
-        XCTAssertNotNil(tokens.get(index: 254))
+        XCTAssertEqual(tokens.len(), UInt(maxTokensCount))
```
kettlebell, line 53: I noticed that we don't have associated tests of this sort for WASM. Probably a good idea to have at least this test included. If you like I can make an issue for it, and I will get around to implementing it.

kettlebell (approved):
Looks good to me! It'd be nice to just have that WASM test for the tokens too.

bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
greenhat: Good catch! Actually, I think `Tokens` type in both C and Wasm layers should not limit how many tokens they hold via `add_token()`. The max token limit exists only in the `ErgoBox` and should be enforced only there. For example, `Tokens` type is used in `BoxSelector` to represent target tokens. There is no limit there. One can select as many tokens as they want, given that they will be placed later into separate output boxes according to the max token limit per box. Which is enforced in `ErgoBoxCandidateBuilder::build()` and in `ErgoBox` itself via `BoundedVec`.
We have https://github.com/ergoplatform/sigma-rust/issues/594 where we can also remove the `Tokens::add_token()` check.

SethDusek (approved):

bindings/ergo-lib-ios/Tests/ErgoLibTests/TokenTests.swift
kettlebell: Thanks, I got it.

PR #599, opened 2022-07-15
Use gte tokens length check

ross-weir:
Doesn't seem possible but just to be safe


greenhat (approved):
Nice catch! Thank you!

PR #605, opened 2022-07-25
Check input coin preservation in TxBuilder, remove min_change_value

### LINKED ISSUE #593: Remove `min_change_value` in `TxBuilder` since it makes loosing change box way too easy
Spotted in #591 and inspired by #590

It's used to filter out change boxes and not include them in the produced transaction. The idea behind it is that in a case that the change value is so small that it should be given to miners because it's smaller than the minimum value allowed for the box to have.
Actually, this will not work as intended since the `ErgoBoxAssets::value` used for representing the change box is of type `BoxValue` that already has the lower bound set to a reasonably minimal box value. So, the change cannot be less the box minimal value anyway.
Besides that, I think that another reason for removal is that `TxBuilder` API should not strive to be smart and manipulate transaction outputs/inputs behind the scenes.

Todo:
- remove `min_change_value`;
- ensure a good error message in the box selector that the change ERG value is less than the minimum box value;

greenhat:
Close #593

Done:
- remove `min_value_change` parameter in `TxBuilder` used to filter out change boxes.
- add check that all input coins cannot be less than all output coins. With existing `inputs < outputs` check we now have only one happy path `inputs == outputs`(coin preservation).


kettlebell (approved):
Looks good to me! Just a silly question: if inputs < outputs, we're effectively burning ERG right?

greenhat:
> Looks good to me! Just a silly question: if inputs < outputs, we're effectively burning ERG right?

Good question! Although we have inputs = outputs check in the protocol - https://github.com/ergoplatform/ergo/blob/66ae8ad214e0f56bd8ad39bda0c752fe5a3354a4/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala#L396
So such tx should simply be rejected. 
I believe doing an early check here is a good idea.

kettlebell:
Thanks! I agree.

PR #606, opened 2022-07-26
Remove undesired error-silencing `Result` to `Option` conversions

### LINKED ISSUE #592: Review all `Result` to `Option` conversion in the codebase for silencing errors
Spotted in #590 and #591

As the cause of #590 the following code:
https://github.com/ergoplatform/sigma-rust/blob/26155df207dc88601de1ae06a5bf5b049d8dcabf/ergo-lib/src/wallet/box_selector/simple.rs#L157-L158

shows the danger of converting `Result` to `Option` where the target value is `Option`. 

Keep in mind that such conversion is perfectly safe where the `Option` on the target describes the error case (`None`).

greenhat:
Close #592 

In this PR, I removed every unwanted error-silencing `Result` to `Option` conversion that I could find. Most of them are related to `Vec<Token>` -> `Option<BoundedVec<Tokens>>` conversion.

### DIFF
--- a/ergo-lib/src/chain/transaction/unsigned.rs
+++ b/ergo-lib/src/chain/transaction/unsigned.rs
@@ -9,7 +9,6 @@ use super::Transaction;
 use super::TxIoVec;
 use super::{distinct_token_ids, TransactionError};
 use bounded_vec::BoundedVec;
-use bounded_vec::OptBoundedVecToVec;
 use ergo_chain_types::blake2b256_hash;
 use ergotree_interpreter::sigma_protocol::prover::ProofBytes;
 use ergotree_ir::chain::ergo_box::ErgoBoxCandidate;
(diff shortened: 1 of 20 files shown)


ergo-lib/src/chain/transaction/unsigned.rs
```diff
@@ -69,7 +81,7 @@ impl UnsignedTransaction {
         Ok(TxId(blake2b256_hash(&bytes)))
     }
 
-    fn to_tx_wo_proofs(&self) -> Result<Transaction, SigmaSerializationError> {
+    fn to_tx_wo_proofs(&self) -> Transaction {
```
kettlebell, line 84: How about:

```suggestion
    fn to_tx_without_proofs(&self) -> Transaction {
```

kettlebell (changes requested):
Looks good to me! Just one tiny nitpick.

SethDusek (approved):

PR #615, opened 2022-08-16
use js_sys::Error in Wasm error conversion;

greenhat:
Use `js_sys::Error` instead of `String` for errors returned via `Result` in Wasm API.
It seems more fitting since these errors end up thrown (see https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html )


kettlebell (approved):
Looks good to me! First time I've come across the `stringify` macro too, it's pretty nifty.

PR #636, opened 2022-10-02
Fix clippy errors relating to deprecated functions in `wasm-bindgen` and broken SPV test

### LINKED ISSUE #635: Fix clippy errors relating to deprecated functions in `wasm-bindgen`
Shown in this GA run: https://github.com/ergoplatform/sigma-rust/pull/634/checks?check_run_id=8663370162

An example of a deprecation is here: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html#method.from_serde

### LINKED ISSUE #638: Increase minimum ergo node version from 4.0.28 to 4.0.100 for SPV workflow tests
Due to the proposed EIP-37 hard-fork, the SPV tests were breaking since a number of nodes have already upgraded and we use string-comparison for ergo node versions (so that "4.0.100" is judged to be less than "4.0.28"). EIP-37 looks like it will pass, so let's bump up the version to 4.0.100.

kettlebell:
Close #635
Close #638

### DIFF
--- a/bindings/ergo-lib-ios/Tests/ErgoLibTests/RestNodeApiTests.swift
+++ b/bindings/ergo-lib-ios/Tests/ErgoLibTests/RestNodeApiTests.swift
@@ -106,7 +106,7 @@ final class RestNodeApiTests: XCTestCase {
         let txId = try TxId(withString: "258ddfc09b94b8313bca724de44a0d74010cab26de379be845713cc129546b78")
         let proofs = try await withThrowingTaskGroup(of: [NipopowProof].self) { group -> [NipopowProof] in
             group.addTask {
-                let proof = try await getNipopowProof(url: URL(string: "159.65.11.55:9053")!, headerId: headerId)!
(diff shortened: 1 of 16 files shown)


bindings/ergo-lib-ios/Tests/ErgoLibTests/RestNodeApiTests.swift
```diff
@@ -106,7 +106,7 @@ final class RestNodeApiTests: XCTestCase {
         let txId = try TxId(withString: "258ddfc09b94b8313bca724de44a0d74010cab26de379be845713cc129546b78")
         let proofs = try await withThrowingTaskGroup(of: [NipopowProof].self) { group -> [NipopowProof] in
             group.addTask {
-                let proof = try await getNipopowProof(url: URL(string: "159.65.11.55:9053")!, headerId: headerId)!
```
kettlebell, line 109: `159.65.11.55` currently has version < 4.0.100, so switched with a compatible one.

greenhat (approved):
Looking great! I appreciate you taking care of this!

PR #637, opened 2022-10-03
Add {set_exports} script for {ergo-lib-wasm}

### LINKED ISSUE #632: Issue running `ergo-lib-wasm-browser` in Node environments
Although it may seem weird to run `ergo-lib-wasm-browser` in Node environments (as we have `ergo-lib-wasm-node`), it makes sense when we consider writing tests for apps that use `ergo-lib-wasm-browser` using tools running in Node. It's not possible to write tests using Jest or Vitest, for example, if we add this package to an app dependencies.

The issue is not directly related to `sigma-rust`, however; Unfortunately, `wasm-pack` output `package.json` only contains a `module` field. There are no `type: module`, `main` or `exports` fields available so that Node can handle the package as expected. There is [an open pull request](https://github.com/rustwasm/wasm-pack/pull/1061) in `wasm-pack` repo to fix the issue, but it's not merged yet.

Meanwhile, we can wait for `wasm-pack` to fix their issue and use tools such as `patch-package` to bypass the problem, or we can apply some minor modifications to the output of `wasm-pack` and add required fields to the generated `package.json`.

mkermani144:
When {wasm-pack} is run with browser target, it generates a {package.json} in which there is a {module} field, but no {main}, {type} or {exports} field is present. This causes some of the modern tools which run in Node be unable to use the package (Vitest is an example).

Although it may seem irrational to use the browser package in Node, but it makes sense when we want to write tests. Many test tools such as Jest and Vitest run in Node.

Add a {set_exports} script to be run after {wasm-pack}, updating the generated {package.json} as follows:
- Add an {exports} field to enable new Node entry resolution feature
- Add {"type": "module"} to indicate that this is an ESM module
- Remove {module}, as it's no longer needed

{wasm-pack} may fix the issue in the future or new module entry resolution mechanisms may be introduced in the future. Until then, this script will enable the use of the package without getting browser environment related issues when run in Node.

Related {wasm-pack} PR:
https://github.com/rustwasm/wasm-pack/pull/1061

Closes #632.

### DIFF
--- a/bindings/ergo-lib-wasm/package.json
+++ b/bindings/ergo-lib-wasm/package.json
@@ -10,8 +10,8 @@
         "doc": "jsdoc -c jsdoc.json pkg/ergo_wallet_wasm.js README.md -d docs",
         "build-nodejs": "rm -rf ./pkg-nodejs && wasm-pack build --target nodejs --out-dir pkg-nodejs && cd pkg-nodejs && node ../scripts/publish_helper -nodejs",
         "build-nodejs-alpha": "rm -rf ./pkg-nodejs && wasm-pack build --target nodejs --out-dir pkg-nodejs && cd pkg-nodejs && node ../scripts/publish_helper -nodejs && npm version minor && node ../scripts/set_alpha_version -nodejs $(git rev-parse --short HEAD)",
-        "build-browser": "rm -rf ./pkg-browser && wasm-pack build --target browser --out-dir pkg-browser --features rest && cd pkg-browser && node ../scripts/publish_helper -browser",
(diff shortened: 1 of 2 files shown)


greenhat:
Thank you! Any chance these changes brake anyone's browser workflow?

mkermani144:
I'm not sure, as it depends on the package resolution of build tool used. If there is a build tool which doesn't support `exports` field, this becomes a breaking change. But as far as I know, the `exports` field is supported in the well known build tools and we shouldn't count this as a breaking change.
In addition, if someone uses old versions of Node (`~v12` or older, I think), this should be counted as breaking change, too.

greenhat (approved):
Looking great! Thank you!

PR #644, opened 2022-10-25
Fix bottom-up and top-down traverals in prover

### LINKED ISSUE #579: Check polynomial in Cthreshold* has correct size
### Motivation
`polynomial` in `CthresholdUnproven` and `CthresholdUnchecked` when [parsed]( https://github.com/ergoplatform/sigma-rust/blob/a7dee7a013662677771385d228931efe5215d25d/ergotree-interpreter/src/sigma_protocol/sig_serializer.rs#L219-L222 ) in `sig_serialized` is expected to be `(ct.children.len() - ct.k) * SOUNDNESS_BYTES` bytes long. 
To ensure proper parsing we should check that `polynomial` has the correct size. 

### Implementation details
I believe a failable ctor is the best way to ensure `Cthreshold*` correctness, i.e. fail early. Thus, forcing the call site to handle the possible error. Also, `with_polynomial` method should be updated as well. The check should be similar to the temporary assert I put in `sig_serializer` - https://github.com/ergoplatform/sigma-rust/blob/a7dee7a013662677771385d228931efe5215d25d/ergotree-interpreter/src/sigma_protocol/sig_serializer.rs#L96-L101

### LINKED ISSUE #643: Sync top-down and bottom-up traversal in the prover with Scala version
rewrite function at https://github.com/ergoplatform/sigma-rust/blob/7f44eceb3ebd0b008551437174d78f925493a927/ergotree-interpreter/src/sigma_protocol/proof_tree.rs#L147-L241
even though it says top-down in the comment, it is bottom-up. See how children nodes are traversed before the parent(current) node.

It might help with https://github.com/ergoplatform/sigma-rust/issues/579#issuecomment-1259058014

greenhat:
Close #643
Close #579

Apply @kettlebell fix https://gist.github.com/kettlebell/21f94906a89072afb97fbfe5ff78df33 to `rewrite` and rename it to `rewrite_bu`.
Rename original `rewrite` to `rewrite_td`.
Add `test_prove_inner_threshold` from https://github.com/ergoplatform/sigma-rust/issues/579#issuecomment-1259058014 to simulate fail on "wrong" traversal order (BU vs TD).
Make `CthresholdUnproven::with_polynomial` failable (check polynomial bytes size against children count and k).

### DIFF
--- a/ergotree-interpreter/src/sigma_protocol/proof_tree.rs
+++ b/ergotree-interpreter/src/sigma_protocol/proof_tree.rs
@@ -144,9 +144,106 @@ pub(crate) enum ProofTreeKind<'a> {
     Conjecture(&'a dyn ProofTreeConjecture),
 }
 
+/// Traverses the tree in the bottom-up manner, calling `f` for every node/leaf and setting
+/// it's returned value (if `Some`) as new node/leaf or do nothing if it's returned `None`
+pub(crate) fn rewrite_bu<F>(tree: ProofTree, f: &F) -> Result<ProofTree, ProverError>
+where
+    F: Fn(&ProofTree) -> Result<Option<ProofTree>, ProverError>,
+{
+    let cast_to_ust = |children: SigmaConjectureItems<ProofTree>| {
+        children.try_mapped(|c| {
+            if let ProofTree::UncheckedTree(ust) = c {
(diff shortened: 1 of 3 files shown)


kettlebell (approved):
Looks great, thanks!

PR #658, opened 2022-12-04
Digest is now Copy as well as all major wrappers based on it.

oskin1:


greenhat (approved):
Looking great! Thank you!

PR #662, opened 2022-12-11
Replace instances of `Rc<ErgoBox>` with `Arc<ErgoBox>`

### LINKED ISSUE #661: Replace instances of `Rc<ErgoBox>` with `Box<ErgoBox>`
Allows `Transaction` and other types to be `Send`, so we can pass it between threads.

kettlebell:
Close #661

### DIFF
--- a/ergo-lib/src/wallet/multi_sig.rs
+++ b/ergo-lib/src/wallet/multi_sig.rs
@@ -293,7 +293,7 @@ pub fn extract_hints(
             .enumerate()
             .map(|(idx, b)| ErgoBox::from_box_candidate(b, tx_ctx.spending_tx.id(), idx as u16))
             .collect::<Result<Vec<ErgoBox>, SigmaSerializationError>>()?;
-        let outputs_ir = outputs.into_iter().map(Rc::new).collect();
+        let outputs_ir = outputs.into_iter().map(Box::new).collect();
         let data_inputs_ir = if let Some(data_inputs) = tx_ctx.spending_tx.data_inputs.as_ref() {
             Some(data_inputs.clone().enumerated().try_mapped(|(idx, di)| {
                 tx_ctx
@@ -302,7 +302,7 @@ pub fn extract_hints(
                     .ok_or(TxSigningError::DataInputBoxNotFound(idx))?
                     .iter()
                     .find(|b| di.box_id == b.box_id())
-                    .map(|b| Rc::new(b.clone()))
+                    .map(|b| Box::new(b.clone()))
                     .ok_or(TxSigningError::DataInputBoxNotFound(idx))
             })?)
         } else {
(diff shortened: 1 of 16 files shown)


oskin1 (approved):

greenhat (changes requested):
Looking good! I'm thrilled to see sigma-rust in a multi-threaded environment!
Have you considered thread-safe `Arc` instead of `Box`? I recall the reason for `Rc` was that ergo boxes have quite a big memory footprint, so I aimed to eventually eliminate all `.clone()` calls and use references as much as possible for ergo boxes. Although `Box` is also a reference, but only for single ownership, which I'm feeling is not enough for ergo boxes (think txs, context in the interpreter, etc.).

kettlebell:
> Looking good! I'm thrilled to see sigma-rust in a multi-threaded environment! Have you considered thread-safe `Arc` instead of `Box`? I recall the reason for `Rc` was that ergo boxes have quite a big memory footprint, so I aimed to eventually eliminate all `.clone()` calls and use references as much as possible for ergo boxes. Although `Box` is also a reference, but only for single ownership, which I'm feeling is not enough for ergo boxes (think txs, context in the interpreter, etc.).

Yes, I did consider it. I had a look at how `Rc<ErgoBox>` was used throughout the codebase and nothing jumped out at me in terms of shared ownership. I chose to use `Box` because I know that `Arc` imposes extra runtime cost over `Rc`. I'm happy to change it over to `Arc`, no problem at all.

greenhat:
> Yes, I did consider it. I had a look at how `Rc<ErgoBox>` was used throughout the codebase and nothing jumped out at me in terms of shared ownership. I chose to use `Box` because I know that `Arc` imposes extra runtime cost over `Rc`. I'm happy to change it over to `Arc`, no problem at all.

Thank you! My bad. There are way too many `.clone()` calls still there but I hope to get rid of them eventually.

greenhat (approved):
Thank you!

oskin1:
@greenhat isn't adding Arc inside ErgoBox too opinionated solution? Whoever wanting to share ownership of ErgoBox can wrap it into Arc in whole.

greenhat:
> @greenhat isn't adding Arc inside ErgoBox too opinionated solution? Those who want to share ownership of ErgoBox can wrap it into Arc in whole.

I'm not sure I understand what you mean by "Arc inside ErgoBox". 
The `Rc` (changed in this PR) was used in the interpreter `Context` type. In the future, I'd like not to clone ErgoBox but pass a reference and share the ownership between tx, context, etc.

greenhat:
My aim is to share the ownership to the ref to the ErgoBox. The ErgoBox itself owns it's data and there is no references used for it's internal parts.

oskin1:
@greenhat My bad, I misunderstood how CBox relates to ErgoBox.

PR #669, opened 2023-01-09
expose constant value and type in c bindings

Alesfatalis:
Makes the value and type of a Constant available as debug string in the bindings for C


greenhat (approved):
Looking great! Welcome, and thank you for your contribution!

PR #678, opened 2023-01-27
Add `FullBlock` to `ergo-lib`

kettlebell:


greenhat (changes requested):
Looking good! Thank you! 
Would I be able to parse it from JSON? I'm seeing `blockTransactions` field name in JSON, while here it's `block_transactions`. 

Could you please fix the build wo default features?

greenhat:
Oh, and I fixed the test coverage CI job in develop. Please, rebase.

kettlebell:
> Looking good! Thank you! Would I be able to parse it from JSON? I'm seeing `blockTransactions` field name in JSON, while here it's `block_transactions`.
> 
> Could you please fix the build wo default features?

Thanks for the find, I've fixed the field name. It slipped through because we don't use `blocks/{headerId}` endpoint directly to create `FullBlock`. All issues fixed plus added JSON roundtrip test.

greenhat (approved):
Looking great! Thank you, I appreciate the added test!

PR #689, opened 2023-02-10
Introduce Wallet::sign_tx_input;

greenhat:


capt-nemo429 (commented):
That's exactly what I need, thanks!

PR #721, opened 2023-08-29
`EcPoint` derive from `ProjectivePoint` + update `k256` dependency

kettlebell:


greenhat (approved):
Looking great! Thank you!

PR #728, opened 2023-10-12
Reset interpreter env after lambda call

### LINKED ISSUE #730: `Env` needs resetting after `Lambda` calls and `BlockValue`
A change that was introduced in v0.25 was in the `Evaluable` trait where `fn eval` now takes a mutable reference to `Env`. This introduces errors when lambdas are called and `BlockValues` are evaluated. Consider the diff between v0.24 and v0.25 for `Coll.exists`:
image

We used to clone the `Env`, add the arg, then evaluate the function body. Now with mutable `Env`, the arg persists in the `Env` instance for subsequent `eval` calls. This will break validation since lambas are built by the interpreter like
```
val v15 = {
        (v15: Coll[Byte]) => 
          if (upcast(v15(0)) < 0) Coll[Coll[Byte]](, 0).append(v15) else v15
        }
```
See how `v15` is reused as an arg for the lambda which is already associated with `v15`. A subsequent use of `v15` will error because it will try to `Apply` on a `Coll[Byte]`.

------
I ran across this issue when trying to validate this [contract](https://github.com/spectrum-finance/spectrum/blob/b04ea4834635a15c28a37a4f18e6f27f5a714e83/spectrum-ergo-connector/contracts/signature_verification.sc#L254) which makes heavy use of lambdas. It validated perfectly for v0.24.1 and with the `sigmastate-interpreter`. PR #728 will resolve it.

kettlebell:
Close #730.

### DIFF
--- a/ergotree-interpreter/src/eval/block.rs
+++ b/ergotree-interpreter/src/eval/block.rs
@@ -12,31 +12,48 @@ use crate::eval::Evaluable;
 
 impl Evaluable for BlockValue {
     fn eval(&self, env: &mut Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
-        let mut existing_variables = HashMap::new();
-        let mut new_variables = vec![];
-        for i in &self.items {
-            // TODO: new try_extract_spanned_into?
-            let spanned_val_def = &i.clone().try_extract_into::<Spanned<ValDef>>()?;
-            let val_def = spanned_val_def.expr();
-            let idx = val_def.id;
-            let v: Value = val_def.rhs.eval(env, ctx)?;
-            if let Some(old_val) = env.get(idx) {
-                existing_variables.insert(idx, old_val.clone());
-            } else {
-                new_variables.push(idx);
+        // The start of the top-level block of statements does not contain any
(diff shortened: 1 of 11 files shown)


ergotree-interpreter/src/eval/block.rs
```diff
@@ -11,14 +12,31 @@ use crate::eval::Evaluable;
 
 impl Evaluable for BlockValue {
     fn eval(&self, env: &mut Env, ctx: &mut EvalContext) -> Result<Value, EvalError> {
+        let mut existing_variables = HashMap::new();
```
kettlebell, line 15: This code breaks the `diag_on_reduced_to_false` test. We might need to make an exception for the global `BlockValue` of a script, and do not delete introduced `ValId`s?

greenhat: Thank you for taking care of this! Ideally, yes. If not, feel free to make the test ignored/commented.

kettlebell: No problem! Made the change so `ValDef`s in global block persist.

greenhat (approved):
Looking great! Thank you!

PR #735, opened 2023-12-22
Get min box value without checking value is valid

### LINKED ISSUE #734: Allow box min value & size to be calculated without value validation
> We have a problem using builder.calc_min_box_value() in box builders, since its trying to first build the box and then calculate the min value. We get this error: ErgoBoxCandidateBuilderError: Box value BoxValue(1100000) is too low, minimum value for box size of 3392 bytes is: BoxValue(1221120) nanoERGs .
Can u update the implemetation somehow, not to build the box for this calculations? We have another option to use a large value as the box value and update that amount after this calculation, but we rather not to use such approach.

We could have a method that builds the box but doesn't validate the size/value, this will allow the min box value to be calculated and set accordingly afterwards

ross-weir:
closes #734

### DIFF
--- a/ergo-lib/src/chain/ergo_box/box_builder.rs
+++ b/ergo-lib/src/chain/ergo_box/box_builder.rs
@@ -143,10 +143,7 @@ impl ErgoBoxCandidateBuilder {
 
         // Won't be overflowing an i64, so unwrap is safe.
         #[allow(clippy::unwrap_used)]
-        Ok(
-            BoxValue::try_from(bytes.len() as i64 * BoxValue::MIN_VALUE_PER_BOX_BYTE as i64)
-                .unwrap(),
-        )
+        Ok(BoxValue::try_from(bytes.len() as i64 * self.min_value_per_byte as i64).unwrap())
     }
 
     /// Set register with a given id (R4-R9) to the given value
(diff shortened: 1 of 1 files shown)


ergo-lib/src/chain/ergo_box/box_builder.rs
```diff
@@ -135,6 +135,17 @@ impl ErgoBoxCandidateBuilder {
         )
     }
 
+    /// Calculate minimal box value for the current box serialized size(in bytes) without checking
+    /// if it contains a valid amount of nanoergs.
+    pub fn calc_min_box_value_unchecked(&self) -> Result<BoxValue, ErgoBoxCandidateBuilderError> {
```
sethdusek: Is there any case where developers actually want/need the semantics of calc_min_box_value (i.e. error if box value is below min box value while calculating min box value)? 

In my opinion it'd be better to replace calc_min_box_value with this calc_min_box_value_unchecked since we shouldn't be throwing box validation errors like these before build() is called.

ross-weir: Wasn't sure which is why I was hesitant to change it

I agree & given the library is still 0.1.x version I think it's OK to change the public API in this case

Updated, thanks!

sethdusek (approved):

sethdusek:
LGTM, feel free to merge

PR #736, opened 2023-12-28
Transaction Validation

### LINKED ISSUE #703: Tx verification
We have `verify_tx_input_proof,` and we're missing the whole tx verification.

sethdusek:
Initial stateless validation work (things we can validate without needing blockchain context). Also includes a fix where we were allowing as many as u16::MAX inputs/outputs (65535) when ergo only allows i16::MAX (Short.MaxValue in scala) which is (32767). https://github.com/ergoplatform/ergo/blob/48239ef98ced06617dc21a0eee5670235e362933/ergo-core/src/main/scala/org/ergoplatform/modifiers/mempool/ErgoTransaction.scala#L93


sethdusek:
Should be good to go now! Some notes:

- This PR despite the usual unit tests was tested using [ergovalidation](https://github.com/SethDusek/ergovalidation) which downloads transactions from node (requires extra indexing to be enabled) and validates them. After fixing some script bugs we're now down to 15 transactions on mainnet that sigma-rust rejects, all of which rely on buggy behavior from v4.0 that no longer holds in v5.0, including CONTEXT.selfBoxIndex being -1 and sigmaProp(sigmaProp(true)) being allowed. I believe fixing these (and adding version-specific actions to ergotree-interpreter) is out of scope for this PR. [Gist of failing transactions](https://gist.github.com/SethDusek/9c28ad21738c1513f227bdbac4ad4f6b)

- I had to bump rust version since sigma-rust no longer builds as some dependencies don't support rust 1.64 anymore. Clippy rejection seems to be related and should be ignored for now

- There are some limitations still in interpreter. The biggest one is that atLeast(0, ...) isn't supported (I plan on addressing this separately in another PR, but I fixed atLeast(1, ...) not being supported in this PR). 
- There are some breaking changes in this PR, namely ErgoStateContext now also requiring Parameters. I believe this is an acceptable trade-off even if parameters aren't needed for signing. Later when sigma-rust adds support for costing we could use Parameters to calculate costs during signing
- Also fixed quadratic scaling in TransactionContext::new
old:
txcontextold
new:
txcontextnew

ergotree-interpreter/src/eval/scoll.rs
```diff
@@ -24,15 +24,20 @@ pub(crate) static INDEX_OF_EVAL_FN: EvalFn = |_env, _ctx, obj, args| {
             .get(0)
             .cloned()
             .ok_or_else(|| EvalError::NotFound("indexOf: missing first arg".to_string()))?;
-        let fallback_index = args
```
sethdusek, line 27: indexOf now has the same semantics as scala version, https://github.com/ScorexFoundation/sigmastate-interpreter/blob/efa68e7079393d70d25d43dd63dfba9a18d03415/data/shared/src/main/scala/sigma/ast/methods.scala#L880

ergo-lib/src/chain/transaction/ergo_transaction.rs
```diff
@@ -0,0 +1,162 @@
...
+    InvalidHeightError(u32),
+    #[error("Creation height {0} <= input box max height{1}")]
+    /// After Block V3, all output boxes height must be >= max(inputs.height). See <https://github.com/ergoplatform/eips/blob/master/eip-0039.md> for more information
+    MonotonicHeightError(u32, u32),
+    #[error("Output box's creation height is negative (not allowed after block version 1)")]
+    /// Negative heights are not allowed after block v1.
+    /// When using sigma-rust where heights are always unsigned, this error may be because creation height was set to be >= 2147483648
+    NegativeHeight,
+    #[error("Output box size {0} > maximum {}", ErgoBox::MAX_BOX_SIZE)]
+    /// Box size is > [ErgoBox::MAX_SCRIPT_SIZE]
```
kushti: MAX_BOX_SIZE

kushti (approved):

ergo-lib/src/chain/transaction/ergo_transaction.rs
kushti: Please fix linter issues

kushti:
Please fix linter issues

PR #742, opened 2024-05-02
i741 expose missing types and functions

### LINKED ISSUE #741: Add missing functions/implementations of ExtSecretKey, ExtPubKey and DerivationPath in c-bindings
Current c-bindings implementation is missing some useful functions for `ExtSecretKey` like:

- get derivation path from `ExtSecretKey`
- get secret key from `ExtSecretKey`
- get public key from `ExtSecretKey`
- derive a new `ExtSerectKey` based on derivation path

Currently the function `ext_secret_key_new` in `bindings/ergo-lib-c-core/src/ext_secret_key.rs` accepts a string as derivation path but should be `DerivationPath` which needs to be implemented first since it is missing (see below).

Following types are completely missing in c-bindings:

- `ExtPubKey`
- `DerivationPath`

Alesfatalis:
closes #741

### DIFF
--- a/bindings/ergo-lib-ios/Sources/ErgoLib/ExtSecretKey.swift
+++ b/bindings/ergo-lib-ios/Sources/ErgoLib/ExtSecretKey.swift
@@ -5,12 +5,9 @@ class ExtSecretKey {
     internal var pointer: ExtSecretKeyPtr
 
     /// Create ExtSecretKey from secret key bytes, chain code and derivation path
-    /// Derivation path should be a string in the form of: m/44/429/acc'/0/addr
-    init(secretKeyBytes: [UInt8], chainCodeBytes: [UInt8], derivationPathStr: String) throws {
+    init(secretKeyBytes: [UInt8], chainCodeBytes: [UInt8], derivationPath: DerivationPath) throws {
         var ptr: ExtSecretKeyPtr?
-        let error = derivationPathStr.withCString { cs in
-            ergo_lib_ext_secret_key_new(secretKeyBytes, chainCodeBytes, cs, &ptr)
-        }
+        let error = ergo_lib_ext_secret_key_new(secretKeyBytes, chainCodeBytes, derivationPath.pointer, &ptr)
         try checkError(error)
         self.pointer = ptr!
     }
--- a/bindings/ergo-lib-c-core/src/derivation_path.rs
+++ b/bindings/ergo-lib-c-core/src/derivation_path.rs
@@ -1,0 +1,76 @@
+//! Derivation Path functionality
+
+use crate::util::const_ptr_as_ref;
+use crate::{util::mut_ptr_as_mut, Error};
+use derive_more::{From, Into};
+use ergo_lib::wallet;
+use ergo_lib::wallet::derivation_path::{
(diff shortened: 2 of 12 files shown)


sethdusek:
CI issues should be fixed now, try rebase

bindings/ergo-lib-ios/Sources/ErgoLib/ExtSecretKey.swift
```diff
@@ -5,12 +5,9 @@ class ExtSecretKey {
...
 
     /// Create ExtSecretKey from secret key bytes, chain code and derivation path
-    /// Derivation path should be a string in the form of: m/44/429/acc'/0/addr
-    init(secretKeyBytes: [UInt8], chainCodeBytes: [UInt8], derivationPathStr: String) throws {
+    init(secretKeyBytes: [UInt8], chainCodeBytes: [UInt8], derivationPath: DerivationPath) throws {
         var ptr: ExtSecretKeyPtr?
-        let error = derivationPathStr.withCString { cs in
-            ergo_lib_ext_secret_key_new(secretKeyBytes, chainCodeBytes, cs, &ptr)
-        }
+        let error = ergo_lib_ext_secret_key_new(secretKeyBytes, chainCodeBytes, derivationPath.pointer, &ptr)
```
sethdusek, line 10: It is possible to cause UB from Swift (or any other language using C bindings) if secretKeyBytes < SecretKeyBytes::LEN or chaincodeBytes < chainCode::LEN when creating a slice

Swift/Go bindings should check if this is the case and error sanely. For the C bindings themselves it'd be sufficient to mention in the documentation that secret_key_bytes needs to be SecretKeyBytes::LEN bytes

sethdusek (changes requested):

bindings/ergo-lib-ios/Sources/ErgoLib/ExtSecretKey.swift
Alesfatalis: @SethDusek Agreed with adding length of the bytes to documentation for the C bindings(will add them). In regards to Swift bindings for the time being this isn't handled anywhere in Swift bindings where bytes are passed and should probably tackled in a separate issue improving the bindings? See for example:
- `MerkleProof.swift`
- `SecretKey.swift`

sethdusek: MerkleProof.swift does handle it correctly by passing the length of the byte array to any rust functions (which later gets validated by things like Digest32::try_from).

But this issue does predate your PR even in ExtSecretKey.swift so I suppose this should be handled in a separate issue.

https://github.com/sigmaspace-io/ergo-lib-go/blob/main/secretkey.go#L39 I've also discovered a similar issue in the go bindings. I can create an issue there as well if you'd like

sethdusek (approved):

bindings/ergo-lib-ios/Sources/ErgoLib/ExtSecretKey.swift
Alesfatalis: Thank you for noticing already fixed for ergo-lib-go on develop branch: https://github.com/sigmaspace-io/ergo-lib-go/commit/107926abb4f491f8594a3cd38fdc4de8b1952366

PR #747, opened 2024-05-13
fix storage fee calculation

### LINKED ISSUE #746: Bug: Storage fee calculation uses output candidate instead of input box for storage fee calculation
Storage fee calculation uses output candidate instead of input box for storage fee calculation:
https://github.com/ergoplatform/sigma-rust/blob/1801136a167c7f83b63a68ce315793daccca6191/ergo-lib/src/chain/transaction/storage_rent.rs#L38-L39

According to:
https://github.com/ergoplatform/ergo/blob/1a417f44459e78acdb4edf6cc00f6d594494faf0/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/ErgoInterpreter.scala#L34-L45

input box should be used to calculate if storage fee exceeds the box value

Alesfatalis:
closes #746

### DIFF
--- a/ergo-lib/src/chain/transaction/storage_rent.rs
+++ b/ergo-lib/src/chain/transaction/storage_rent.rs
@@ -35,7 +35,7 @@ pub(crate) fn try_spend_storage_rent(
             .ok()?;
         let output_candidate = context.outputs.get(output_idx as usize)?;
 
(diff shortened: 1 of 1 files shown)


sethdusek (approved):
Good catch, thank you!

PR #795, opened 2024-12-15
Global frombytearray

sebastianrof:


sebastianrof:
@SethDusek i fixed the clippy and rustfmt errors in the workflow

ergotree-ir/src/types/sglobal.rs
```diff
@@ -61,3 +64,18 @@ lazy_static! {
...
 }
+
+lazy_static! {
+    static ref FROM_BIGENDIAN_BYTES_METHOD_DESC: SMethodDesc = SMethodDesc {
+        method_id: FROM_BIGENDIAN_BYTES_METHOD_ID,
+        name: "fromBigEndianBytes",
+        tpe: SFunc {
+            t_dom: vec![SType::SGlobal, SType::SColl(SType::SByte.into())],
+            t_range:SType::STypeVar(STypeVar::t()).into(),
+            tpe_params: vec![STypeParam::param_t()],
```
sethdusek: I don't think this is the right way to do this. Since the output type of fromBigEndianBytes can't be inferred from input, the output type needs to be explicitly serialized/deserialized, so add it to explicit_type_args list, see: https://github.com/ergoplatform/sigma-rust/blob/f6048c1e4e4813af9308c484b3ab7fe904ce7005/ergotree-ir/src/types/sbox.rs#L60

Also, can you add roundtrip tests for MethodCalls using this new method? You can skip versioning for now as versioning semantics are still not final in sigmastate-interpreter

ergotree-interpreter/src/eval/sglobal.rs
```diff
@@ -53,18 +54,115 @@ pub(crate) static XOR_EVAL_FN: EvalFn = |_mc, _env, _ctx, obj, args| {
...
+            Ok(Value::Long(i64::from_be_bytes(long_bytes)))
+        }
+        SType::SBigInt => {
+            if bytes.len() > 32 {
+                return Err(EvalError::UnexpectedValue(
+                    "BigInt value doesn't fit into 32 bytes in fromBigEndianBytes".to_string(),
+                ));
+            }
+            let bytes_vec: Vec<u8> = bytes.iter().map(|&x| x as u8).collect();
+            let big_int = num_bigint::BigInt::from_bytes_be(num_bigint::Sign::Plus, &bytes_vec);
```
sethdusek: This would make it impossible to deserialize negative bigints, I suggest using BigInt256::from_be_slice instead

ergotree-interpreter/src/eval/sglobal.rs
```diff
@@ -53,18 +54,115 @@ pub(crate) static XOR_EVAL_FN: EvalFn = |_mc, _env, _ctx, obj, args| {
...
+            let b1 = bytes[1] as i16;
+            Ok(Value::Short(((b0 & 0xFF) << 8 | (b1 & 0xFF)) as i16))
+        }
+        SType::SInt => {
+            if bytes.len() != 4 {
+                return Err(EvalError::UnexpectedValue(
+                    "To deserialize Int with fromBigEndianBytes, exactly four bytes should be provided".to_string(),
+                ));
+            }
+            let bytes_vec: Vec<u8> = bytes.iter().map(|&x| x as u8).collect();
```
sethdusek: Better to do this without heap allocating, rust doesn't have any great ways to collect into fixed-size arrays, so maybe a fold that accumulates bytes would be better: something like bytes.iter().map(|b| b as u8).fold(0i32, |acc, x| (acc << 8) || x as i32)

ergotree-interpreter/src/eval/sglobal.rs
```diff
@@ -98,4 +196,98 @@ mod tests {
             expected_xor.as_slice()
         );
     }
+
+    #[test]
```
sethdusek: Can you add roundtrip tests (for example i64->big-endian byte array->back to i64 using this method) for arbitrary bytes/shorts/etc using proptest? See other proptest examples like in src/eval/downcast.rs

ergotree-ir/src/types/stype_param.rs
```diff
@@ -100,3 +100,14 @@ pub struct STypeParam {
     upper_bound: Option<SType>,
     lower_bound: Option<SType>,
 }
+
+impl STypeParam {
```
sethdusek: Don't think this is needed, check comments in sglobal.rs

ergotree-interpreter/src/eval/sglobal.rs
```diff
@@ -98,4 +196,98 @@ mod tests {
...
+        let expr: Expr = MethodCall::new(
+            Expr::Global,
+            sglobal::FROM_BIGENDIAN_BYTES_METHOD
+                .clone()
+                .with_concrete_types(&type_args),
+            vec![bytes.into()],
+        )
+        .unwrap()
+        .into();
+        let ctx = force_any_val::<Context>();
```
sethdusek: Consider using eval_out_wo_ctx instead for clarity

SethDusek (changes requested):
Hi, thanks for the contribution. Please see my comments

ergotree-ir/src/types/sglobal.rs
sebastianrof: Thanks for the review. I'll add explicit type args for proper type inference and include proptest roundtrip tests for MethodCall.

ergotree-interpreter/src/eval/sglobal.rs
sebastianrof: Using BigInt256::from_be_slice now to handle negative bigints correctly.

ergotree-interpreter/src/eval/sglobal.rs
sebastianrof: Fixed to use fold for heap-free byte accumulation.

ergotree-interpreter/src/eval/sglobal.rs
sebastianrof: Added proptest roundtrip tests following the pattern in downcast.rs.

ergotree-ir/src/types/stype_param.rs
sebastianrof: you are right. i removed it.

ergotree-interpreter/src/eval/sglobal.rs
sebastianrof: Done.

sebastianrof:
@SethDusek, I've made the requested changes. Could you please review again?

ergotree-interpreter/src/eval/sglobal.rs
```diff
@@ -53,18 +54,115 @@ pub(crate) static XOR_EVAL_FN: EvalFn = |_mc, _env, _ctx, obj, args| {
...
+                ));
+            }
+            let bytes_vec: Vec<u8> = bytes.iter().map(|&x| x as u8).collect();
+            Ok(Value::BigInt(
+                BigInt256::from_be_slice(&bytes_vec).ok_or_else(|| {
+                    EvalError::UnexpectedValue("Failed to convert to BigInt256".to_string())
+                })?,
+            ))
+        }
+        SType::SUnit => {
```
sethdusek: The original implementation does not have a case for SUnit

sebastianrof:
@SethDusek i fix the `build without default features` issue faced in the pipeline.

SethDusek (changes requested):

ergotree-interpreter/src/eval/sglobal.rs
sebastianrof: Yes, Removed it.

sebastianrof:
@SethDusek I removed the `SUint` additional type

SethDusek (approved):
Looks good, but can you please squash your last 4 commits into one?

sebastianrof:
> Looks good, but can you please squash your last 4 commits into one?

Done.

PR #822, opened 2025-04-07
fix: updated the transactionhintsbag implementation

### LINKED ISSUE #817: Fix TransactionHintsBag segregation
TransactionHintsBag splits hints into secret (OwnCommitment) and public hints, but this isn't done correctly. Instead TransactionHintsBag.public_hints contains all CommitmentHints (including OwnCommitment) and  .secret_hints contains SecretProven.

In ergo-lib-python there is also a without_secrets function in HintsBag that removes OwnCommitment which is suitable for sharing with other signers. Consider adding this to the other bindings and native library to prevent API misuse (signers accidentally sharing private randomness leading to private key leakage if they're not careful)

Correct implementation:

https://github.com/ergoplatform/ergo/blob/a74d8f46f9c691bb54429aa1ea79fc20b4d7a57b/ergo-wallet/src/main/scala/org/ergoplatform/wallet/interpreter/TransactionHintsBag.scala#L13-L16

DioBr4nd0:
closes #817 ,

### DIFF
--- a/ergo-lib/src/wallet.rs
+++ b/ergo-lib/src/wallet.rs
@@ -16,7 +16,7 @@ pub mod signing;
 pub mod tx_builder;
 pub mod tx_context;
 
-use crate::ergotree_interpreter::sigma_protocol::prover::hint::{ Hint, HintsBag };
+use crate::ergotree_interpreter::sigma_protocol::prover::hint::{Hint, HintsBag};
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use ergotree_interpreter::sigma_protocol::private_input::PrivateInput;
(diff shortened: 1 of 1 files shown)


ergo-lib/src/wallet.rs
```diff
@@ -284,64 +284,42 @@ impl TransactionHintsBag {
...
+        let (secret, public): (Vec<Hint>, Vec<Hint>) = hints_bag.hints
             .into_iter()
-            .filter(|hint| matches!(hint, Hint::CommitmentHint(_)))
-            .collect();
-        let secret: Vec<Hint> = hints_bag
-            .hints
-            .into_iter()
-            .filter(|hint| matches!(hint, Hint::SecretProven(_)))
-            .collect();
+            .partition(|hint| matches!(hint, Hint::CommitmentHint(_)));
```
sethdusek, line 290: Not all CommitmentHints are secret, only OwnCommitment is

sethdusek (changes requested):
Hi, thanks for the PR! Please check review comments, and fix failing CI tests (rustfmt)

PR #831, opened 2025-06-08
fix:updated the transactionhintsbag implemntation for public and private hints

### LINKED ISSUE #817: Fix TransactionHintsBag segregation

DioBr4nd0:
closes #817 

Hii @sethdusek , Sorry for the wait, I was involved in some launch tasks, and this slipped from my mind, Here's the correct implmentation of the transactionhintshag, Kindly check once if it works.
thanks

### DIFF
--- a/ergo-lib/src/wallet.rs
+++ b/ergo-lib/src/wallet.rs
@@ -282,66 +282,58 @@ impl TransactionHintsBag {
         }
     }
 
+    /// Private helper function to correctly segregate a HintsBag into secret and public parts.
+    /// Secret hints are `SecretProven` and `CommitmentHint::OwnCommitment`. All others are public.
+    fn segregate_hints(hints_bag: HintsBag) -> (HintsBag, HintsBag) {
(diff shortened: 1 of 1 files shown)


sethdusek:
Hi, please rebase your PR; the CI issues should be fixed

DioBr4nd0:
sure

PR #836, opened 2025-08-06
serialize sigmaboolean as base-16 string in ergo-lib-wasm bindings

Alesfatalis:
Uses base-16 serialization and deserialization for sigmaboolean for json representation in ergo-lib-wasm bindings. Same like ErgoTree

### DIFF
--- a/ergo-lib/src/chain/transaction/reduced.rs
+++ b/ergo-lib/src/chain/transaction/reduced.rs
@@ -11,6 +11,8 @@ use ergotree_ir::serialization::SigmaSerializable;
 use ergotree_ir::serialization::SigmaSerializeResult;
 use ergotree_ir::sigma_protocol::sigma_boolean::SigmaBoolean;
 
+use super::unsigned::UnsignedTransaction;
+use super::TxIoVec;
 use crate::chain::ergo_state_context::ErgoStateContext;
 use crate::chain::transaction::Transaction;
 use crate::chain::transaction::UnsignedInput;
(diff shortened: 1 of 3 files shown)


kushti (approved):

sethdusek:
Why not encode the entire ReducedTransaction as base-16? The current SigmaBoolean JSON serialization implementation provides a sort of human-readable format for keys/propositions and is also used in other places (HintsBag). 

My main concern is that this is a breaking change. There doesn't seem to be a standard for serializing ReducedTransactions to JSON but this would still cause an incompatibility with older versions of sigma-rust. sigmastate-interpreter doesn't seem to have any JSON serialization for ReducedTransaction so it doesn't matter if we break the format, but I'd prefer not to.

Alesfatalis:
> Why not encode the entire ReducedTransaction as base-16? The current SigmaBoolean JSON serialization implementation provides a sort of human-readable format for keys/propositions and is also used in other places (HintsBag).
> 
> My main concern is that this is a breaking change. There doesn't seem to be a standard for serializing ReducedTransactions to JSON but this would still cause an incompatibility with older versions of sigma-rust. sigmastate-interpreter doesn't seem to have any JSON serialization for ReducedTransaction so it doesn't matter if we break the format, but I'd prefer not to.

@sethdusek There shouldn't be any incompatibility with older sigma-rust versions regarding json serialization as this feature was added by me in the previous PR: https://github.com/ergoplatform/sigma-rust/pull/834 

The base-16 serialization format was chosen as it matches the serialization format of ergo trees (suggested to me by @arobsn ). See use case in: https://github.com/arobsn/keystone-ergo-js/blob/232e76f87e7d196690958ab28a1ae9b7766cdb8e/src/serializer/reducedTransactionSerializer.ts#L37

Issue #763, opened 2024-08-21
Sigma-rust Signed transaction rejected by node


Luivatra:
From the error message the node produces it appears the transaction id is different from the signed tx produced by sigma-rust.

Signed tx json:
[signed-paideia-tx.json](https://github.com/user-attachments/files/16691636/signed-paideia-tx.json)

Node resposne when submitting:
`{
  "error": 400,
  "reason": "bad.request",
  "detail": "Malformed transaction: Scripts of all transaction inputs should pass verification. f135b1df59f45e40bb5d0885bd3e710d15027e4f6961025e442d5da0db7a7340: #1 => Success((false,403))"
}`

Edit:
Adding the unsigned tx as well:
[unsigned-paideia-tx.json](https://github.com/user-attachments/files/16726355/unsigned-paideia-tx.json)

PR #843, opened 2026-02-06
Fix ContextExtension serialization order

Luivatra:
Root Cause: Issue #763                                                                                                                                                                                                                                  
   
  When a ContextExtension has 5+ entries, the Ergo node (Scala 2.12) serializes them in HAMT hash-based order, not sorted/insertion order. sigma-rust used insertion order for all sizes, causing bytes_to_sign to differ, which makes the TxId different,
   which invalidates the proof signature.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                          
  Threshold behavior:                                                                                                                                                                                                                                     
  - 1-4 entries: Scala uses Map1-Map4 (insertion order) — matches sigma-rust                                                                                                                                                                              
  - 5+ entries: Scala uses HashMap (HAMT iteration order) — diverges from sigma-rust                                                                                                                                                                      
                                                                                                                                                                                                                                                          
  The Fix                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                          
  Single file modified: ergotree-ir/src/chain/context_extension.rs                                                                                                                                                                                        
                                                                                                                                                                                                                                                          
  What was added:                                                                                                                                                                                                                                         
  1. scala_212_improve(hc: i32) -> i32 — Replicates Scala 2.12's hash improvement function used by its HAMT                                                                                                                                               
  2. scala_212_hamt_sort_key(key: u8) -> u64 — Computes a sort key that matches HAMT iteration order (encodes 5-bit slots at each trie level)                                                                                                             
  3. Modified sigma_serialize — For 5+ entries, sorts by HAMT slot order before writing; for <5, preserves insertion order (unchanged behavior)                                                                                                           
                                                                                                                                                                                                                                                          
  Tests added (4):                                                                                                                                                                                                                                        
  - test_scala_212_improve — Verifies level-0 HAMT slots match empirically observed node behavior                                                                                                                                                         
  - test_hamt_sort_order_6_entries — Keys {0,1,2,3,4,5} sort to [0,5,1,2,3,4] matching the node                                                                                                                                                           
  - test_serialize_order_5plus_entries — Verifies serialized bytes have keys in HAMT order                                                                                                                                                                
  - test_serialize_order_4_entries_unchanged — Verifies <5 entries still use insertion order

### DIFF
--- a/ergotree-ir/src/chain/context_extension.rs
+++ b/ergotree-ir/src/chain/context_extension.rs
@@ -40,13 +40,60 @@ impl fmt::Display for ContextExtension {
     }
 }
 
+/// Scala 2.12 immutable.HashMap hash improvement function.
+/// Used to predict the HAMT (Hash Array Mapped Trie) iteration order
+/// that the Ergo node (Scala 2.12) uses for ContextExtension serialization.
+///
+/// The Ergo node's ContextExtension uses `scala.collection.immutable.Map` which,
+/// for 5+ entries, becomes a HashMap with hash-based iteration order. This order
+/// differs from sigma-rust's BTreeMap/IndexMap sorted order, causing bytes_to_sign
+/// divergence and transaction rejection.
+///
+/// See: <https://github.com/scala/scala/blob/v2.12.20/src/library/scala/collection/immutable/HashMap.scala>
+/// See: <https://github.com/ergoplatform/sigma-rust/issues/763>
+fn scala_212_improve(hc: i32) -> i32 {
+    let mut h: i32 = hc.wrapping_add(!(hc.wrapping_shl(9)));
+    h = h ^ (((h as u32) >> 14) as i32);
+    h = h.wrapping_add(h.wrapping_shl(4));
+    h ^ (((h as u32) >> 10) as i32)
+}
+
+/// Compute a sort key that matches Scala 2.12 HashMap's HAMT iteration order.
+/// The HAMT uses 5 bits per level from the improved hash, iterating slots 0-31
+/// at each level. The sort key encodes levels from outermost (most significant)
+/// to innermost (least significant).
+fn scala_212_hamt_sort_key(key: u8) -> u64 {
+    let hash = scala_212_improve(key as i32) as u32;
+    let mut sort_key: u64 = 0;
+    for level in 0..7 {
+        sort_key <<= 5;
+        sort_key |= ((hash >> (level * 5)) & 0x1f) as u64;
+    }
+    sort_key
+}
+
 impl SigmaSerializable for ContextExtension {
     fn sigma_serialize<W: SigmaByteWrite>(&self, w: &mut W) -> SigmaSerializeResult {
         w.put_u8(self.values.len() as u8)?;
-        self.values.iter().try_for_each(|(idx, c)| {
-            w.put_u8(*idx)?;
-            c.sigma_serialize(w)
-        })?;
+        if self.values.len() >= 5 {
+            // For 5+ entries, Scala 2.12 uses HashMap which iterates in HAMT order
+            // (based on hash of keys). We must match this order for bytes_to_sign
+            // compatibility with the Ergo node.
+            // See: https://github.com/ergoplatform/sigma-rust/issues/763
+            let mut entries: alloc::vec::Vec<_> = self.values.iter().collect();
+            entries.sort_by_key(|(&idx, _)| scala_212_hamt_sort_key(idx));
+            for (&idx, c) in entries {
+                w.put_u8(idx)?;
+                c.sigma_serialize(w)?;
+            }
+        } else {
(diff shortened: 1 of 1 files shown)


a-shannon:
I added a small test-only patch PR against this branch: https://github.com/Luivatra/sigma-rust/pull/1

It adds Scala 2.12 ContextExtension golden vectors for 8, 32, and 128 ids, plus an exact serialized-byte vector for the 8-id case. This is meant to make the JVM consensus traversal order explicit in sigma-rust CI for this fix.

Local checks on the patch:
- `cargo fmt --check -p ergotree-ir`: PASS
- `git diff --check`: PASS

I could not run `cargo test` in a fresh clone because the workspace has no Cargo.lock and resolves `core2 = 0.4.0`, which crates.io reports as yanked. I did not change dependency metadata in the test-only patch.

odiseusme:
Independent validation against current develop 08dffec.

I tested 154 transactions from 30 recent blocks and a separate targeted sample of three transactions containing a five-variable context extension. Direct serde_json::from_str::<Transaction>, ID comparison, and binary roundtrip passed for both samples: 154/154 and 3/3. The node-JSON read path is sound.

For each targeted transaction, I rebuilt the affected input with all 120 insertion orders.

Without #843:
- Exactly 1/120 permutations reproduced the node ID.
- For {0,1,2,3,5}, the matching order was [0,5,1,2,3].
- For {0,1,2,3,4}, it was [0,1,2,3,4].

With #843 applied cleanly:
- All 120/120 permutations reproduced the node ID in all three cases.
- scala_212_hamt_sort_key predicted both distinct on-chain key orders.

This is end-to-end mainnet confirmation that #843 matches the Scala 2.12 immutable HashMap traversal used by the node paths implicated in #763.

Focused test result: cargo test -p ergotree-ir --features arbitrary context_extension: 8 passed.

Before merge, I recommend rebasing #843 and incorporating and running @a-shannon's 8/32/128 Scala 2.12 ordering vectors from Luivatra/sigma-rust#1, now that #916 has removed the core2 blocker. One adjustment there: the 128-ID vector should stay a pure HAMT-order test. The JVM serializer caps a ContextExtension at 127 entries, so test_serialize_order_matches_scala_212_golden_vectors as written asserts behaviour the node rejects. An exact-five permutation test over {0,1,2,3,5} would also pin the Map4 -> HashMap transition directly.

Mainnet cases:
- ff8ac18ec3bf1d2c84bbfac7935cbc2341431cfc0a36a854c5dfefa463b62a0a, height 1872675, keys {0,1,2,3,5}
- 06411ea6602ea5bcc218c427ad13464137bb194d6d47a8bc15836d02febd8369, height 1872522, keys {0,1,2,3,5}
- 480097419073e99dd4bc6e768506f9166052de3d177a0ec58eed2c67fd3f2049, height 1872491, keys {0,1,2,3,4}

To reproduce from a node: block ID from /blocks/at/{height}, then /blocks/{blockId}/transactions.

I recommend merging #843 for 0.29.0 after the rebase and test adjustments above. Full transaction fixtures do not appear necessary; the compact ordering vectors plus an exact-five permutation test cover the changed mechanism.

a-shannon:
@odiseusme, both test adjustments are now pushed to [Luivatra/sigma-rust#2](https://github.com/Luivatra/sigma-rust/pull/2), the active successor to #1, at `d190a3fcc`.

The 128-ID vector now tests HAMT ordering only. The new test checks all 120 insertion permutations of `{0,1,2,3,5}` against one literal byte vector in order `[0,5,1,2,3]`. The 8/32 serialization checks and exact eight-entry vector remain.

With [#843](https://github.com/ergoplatform/sigma-rust/pull/843)'s patch and these tests applied to develop `08dffec5`, the focused ContextExtension suite passed 11/11 on Rust 1.91.1. Formatting and focused Clippy passed. A five-to-six threshold mutation fails the new test; restoring five makes it pass. An independent Scala 2.12.20 map fixture also confirmed all ordering vectors and the 120 permutations.

The test PR still targets your current #843 head and contains no dependency changes. These are local results for the composition with current develop; upstream CI still needs validation after #843 is rebased. The complete test diff is available in #2, with the small [review increment here](https://github.com/a-shannon/sigma-rust/compare/1390b691b21666e7b07feece41e3e65fc7ba067d...d190a3fcc56e29b6caca8959e4cca83f041532d7).

kushti (approved):

odiseusme:
Confirmed: d190a3fc applies cleanly to develop at 975bebf3, and the context_extension suite passes 11/11 on Rust 1.91.1. Luivatra/sigma-rust#2 still targets Luivatra's now-merged topic branch, so merging it would not add the tests upstream. Please open the same head as a new PR against ergoplatform/sigma-rust:develop so upstream CI can run and the tests can be merged before 0.29.0.

Issue #770, opened 2024-09-17
Fix downcasting for BigInt


sethdusek:
Currently, BigInt cannot be downcasted to any numeric type (even BigInt itself):
```rust
fn downcast_to_bigint(in_v: Value) -> Result<Value, EvalError> {
    match in_v {
        Value::Byte(v) => Ok(BigInt256::from(v).into()),
        Value::Short(v) => Ok(BigInt256::from(v).into()),
        Value::Int(v) => Ok(BigInt256::from(v).into()),
        Value::Long(v) => Ok(BigInt256::from(v).into()),
        _ => Err(EvalError::UnexpectedValue(format!(
            "Downcast: cannot downcast {0:?} to BigInt",
            in_v
        ))),
    }
}
```

In 6.0, downcasting BigInt will be possible. Implement this in interpreter along with versioning so downcasting only works when v6.0 is activated

PR #849, opened 2026-04-04
fix(interpreter): allow BigInt downcast to numeric types for all ErgoTree versions

mwaddip:
## Summary

Removes the `ErgoTreeVersion::V3` guards from BigInt and UnsignedBigInt
downcast match arms in `downcast_to_long`, `downcast_to_int`,
`downcast_to_short`, and `downcast_to_byte`.

## Problem

The emission contract performs arithmetic that promotes intermediate
results to `BigInt` (e.g. `67500000000`). Downcasting this back to `Long`
is valid — the value fits — but the V3 version guard rejects it on
pre-V3 ErgoTrees. This causes evaluation failure at testnet block 2666
during initial sync. The JVM reference handles this silently.

## Fix

The `wrap_downcast` helper already handles overflow correctly (returns
`EvalError` if the value doesn't fit in the target type). The V3 guards
were an unnecessary restriction that prevented valid downcasts on older
ErgoTree versions. Removing them makes the Rust interpreter match the
JVM behavior.

Also fixes a pre-existing bug in the `to_byte` property test that
compared BigInt bounds against `i16::MIN`/`i16::MAX` instead of the
correct `i8::MIN`/`i8::MAX`.

## Test plan

- [x] All 6 downcast property tests updated and passing
- [x] All 344 ergotree-interpreter tests pass
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean

### DIFF
--- a/ergotree-interpreter/src/eval/downcast.rs
+++ b/ergotree-interpreter/src/eval/downcast.rs
@@ -36,16 +36,14 @@ fn downcast_to_bigint<'a>(in_v: Value<'a>, ctx: &Context<'_>) -> Result<Value<'a
     }
 }
 
-fn downcast_to_long<'a>(in_v: Value<'a>, ctx: &Context<'_>) -> Result<Value<'a>, EvalError> {
+fn downcast_to_long<'a>(in_v: Value<'a>, _ctx: &Context<'_>) -> Result<Value<'a>, EvalError> {
     match in_v {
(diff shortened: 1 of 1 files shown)


mwaddip:
Closing — per #770 this is intentionally version-gated for 6.0 activation. We'll work around it with a checkpoint on our end.

PR #854, opened 2026-04-09
Port JIT costing from sigmastate-interpreter

mwaddip:
Ports the JIT cost model from sigmastate-interpreter to ergotree-interpreter. validate() sums cost across all inputs against the tx cap via a shared Context, closing the gap where #846's per-input check lets a malicious tx distribute cost across sub-cap inputs.

Beware that there's some conflicts with #858, explained in comments below.


mwaddip:
## Note: relationship to #858 (lazy constant resolution)

#858 eliminates the deep clone + tree walk in `reduce_to_crypto` by resolving `ConstPlaceholder` nodes lazily during evaluation. It touches some of the same files as this PR (`context.rs`, `eval/expr.rs`, `eval.rs`) but the two are independent.

Whichever merges first, the other needs a small mechanical rebase:

- **If this merges first** → #858 rebase adds `constants` field alongside `jit_cost`/`jit_cost_limit` in `Context`, and adds `ctx.add_jit_cost(5)?` to the `ConstPlaceholder` handler
- **If #858 merges first** → this PR rebase adds `jit_cost`/`jit_cost_limit` fields alongside `constants` in `Context`

mwaddip:
## Update to the #858 rebase guidance

The integration branch (`ergo-node-integration` @ `50f44685`) now carries both PRs layered together and can serve as the reference for either rebase direction. Pinning the conflict points here since this PR's Phase 9 shifted the shape of the clash vs. the pre-Phase-9 note on #858.

**If #858 merges first → this PR rebases as follows:**

- `Context`: add `jit_cost: Cell<u64>` + `jit_cost_limit: Option<u64>` alongside `constants: Option<&[Constant]>`. The Cell is cloned by `with_constants`, not shared — see below.
- `reduce_to_crypto` (Phase 1): keep #858's dual-path structure (deserialize vs. lazy-constants common path), but add a `cost_before = ctx.jit_cost_value()` snapshot before dispatch, and on the common path sync the accumulator back to the caller's ctx after `inner` returns:
  ```rust
  let ctx_with_c = ctx.with_constants(constants);
  let res = inner(root, &ctx_with_c, cost_before);
  ctx.jit_cost.set(ctx_with_c.jit_cost_value());
  ```
  Without this sync, the per-tx cumulative limit in `validate()` loses every input's cost to #858's clone.
- Phase 8 `trivial_reduce`: extend the `Expr::ConstPlaceholder` arm to resolve via `ctx.constants` (not `cp.resolved`) so segregated P2PK still short-circuits at 50 JIT without depending on Phase 9a.
- **Phase 9a (resolved field) becomes redundant** — #858's `ctx.constants` already resolves placeholders lazily. Drop Phase 9a entirely.
- **Phase 9b reduces to a cost-only change** — keep #858's `Expr::ConstPlaceholder` eval arm, just change `ctx.add_jit_cost(5)` to `ctx.add_jit_cost(1)`. Retune `jit_cost_limit_exceeded` (limit 1→0, since 1 JIT no longer exceeds 1) and `test_validate_enforces_cumulative_jit_cost_across_inputs` (switch the segregated Boolean tree to non-segregated `SigmaProp(TrivialProp(true))` so per-input JIT lands cleanly on 50 via `trivial_reduce`). Keep the `segregated_constants_charge_1_not_5_per_placeholder` regression but adapt it to the `ctx.constants` path.

The integration commit that captures this delta: [`9094f693`](https://github.com/mwaddip/sigma-rust/commit/9094f693) (`fix: charge ConstantPlaceholder at JitCost(1), not 5`).

**If this PR merges first → #858 rebase:** unchanged from the pre-existing note on that PR, with one correction: the ConstPlaceholder handler should charge **1** JitCost (matching Phase 9), not 5.

mwaddip:
Pushed 3 commits aligning per-item JIT costing with the Scala reference at empty/edge inputs:

- **e11d418d** — `add_per_item_jit_cost` chunks now match Scala `PerItemCost.chunks = (n-1)/cs + 1` (signed division). Fixes the n=0 undercharge: an empty `Coll[T]` of a chunkSize≥2 type now pays `base + per_chunk` (the JVM charges this; `addSeqCost` has no empty guard). Every n≥1 is byte-identical to the old `ceil(n/cs)`.
- **66b7ef00** — packed boolean-constant collections now charge the per-`Constant` cost (5/elem) the JVM applies; `ConcreteCollection` itself stays `FixedCost(20)`. empty→20, n→20+5n.
- **9d2f1836** — regression vectors (empty `And` / `Blake2b256`).

Cost-parity smoke harness still 78/78 vs Scala; full suite green.

mwaddip:
Integration note: the mismatched-numeric arithmetic coercion (#869 vs develop — coerce `Plus(Int,Long)` etc. to the wider type instead of rejecting) needs a matching JIT cost here. This PR costs only same-type arith (dispatch on left, no Upcast cost), so once #869 lands and this rebases onto it, mismatched arith would be coerced but undercharged (`Plus(Int,Long)` 25 vs JVM 35, missing the +10 Upcast cost).

Cost-bearing version staged: https://github.com/mwaddip/sigma-rust/tree/jit-cost-arith-coercion (= this PR + the coercion + its cost: per-op cost keyed on the wider type + the Upcast cost, bigint=30 else 10). Verified — 14 sweep vectors bit-exact (value + cost 35/25/60 + accept/reject), interpreter suite + parity smoke green. Fold in at integration, or fork it for a PR.

mwaddip:
# PR 854 update notes — running log

Accumulated notes for the PR description update after all fix-plan phases are done.
Assembled from session work on `jit-costing` starting 2026-04-21.

## Architectural findings

**PR 846 does NOT enforce cumulative `jit_cost_limit` across inputs.** It creates a
fresh `Context` per input in `validate()` and only enforces the per-input limit via
`ctx.jit_cost_limit`. There is no `running_jit > total_limit` check anywhere in PR
846's diff. Its mainnet validation (19,549 txs, zero mismatches) proves cost-number
parity on benign txs — it does not prove S4 is closed.

**PR 854 + Phase 1 is strictly more correct on S4 than PR 846.** After Phase 1,
sigma-rust shares one `Context` across all inputs in `validate()` and enforces
`jit_cost_limit` cumulatively via `ctx.add_jit_cost(...)` → matches Scala's
sum-then-check semantics.

## Known caveats to call out in the final PR description

- **`verify_tx_input_proof` is intentionally retained** even though
  `validate()` stops using it — the WASM binding at
  `bindings/ergo-lib-wasm/src/transaction.rs:430` still delegates to it for external
  callers that need per-input verification without full-tx cost accounting.
- **The Phase 1 regression test
  (`test_validate_enforces_cumulative_jit_cost_across_inputs`) is load-bearing.**
  After Phase 2 introduces `estimate_crypto_cost` and Phase 3 adds
  `INTERPRETER_INIT_COST` + `compute_tx_init_cost`, the 3-input `Const(true)`
  threshold and `max_block_cost = 1` tuning may need tightening to keep the test
  isolated to the cumulative-limit regression signal (not spuriously triggered by
  init+crypto overheads on a 1-input tx).
- **`Context::add_jit_cost` takes `u64`** with saturating addition. Per-opcode
  cost literals continue to work unchanged (integer literals infer as u64
  from the param type); `add_per_item_jit_cost` widens its internal u32
  product to u64 at the final call site. The u32/u64 split introduced
  during Phases 2-3 (separate `add_jit_cost_u64` for the per-tx init cost)
  has been reverted — the API is PR 854's own creation with no pre-existing
  external users, so there was no benefit to carrying two methods.
- **`ADD_TO_ENV_COST` (5 JIT) is charged per lambda invocation**, inside
  each coll op's closure (Map/Filter/Fold/ForAll/Exists), rather than
  pre-charged as `5 × n_items` before the iteration loop. This matches
  Scala's per-invocation semantics and preserves short-circuit
  correctness: ForAll/Exists stop charging env cost when the predicate
  terminates iteration early. PR 846 uses the same per-invocation
  placement.
- **Slice cost is computed from the requested range** `max(0, until - from)`,
  not from the clipped output length. Costing is pre-clipping (based on
  what the script asked for); the returned slice is post-clipping
  (intersection with collection bounds). The difference is only
  observable when `until > input.len()` — such a script pays for the
  unclipped work even though only the in-bounds portion is returned.
  PR 846 uses the same formula.
- **P2PK is priced as a unit, not piecewise.** Trees whose proposition
  reduces to a plain SigmaProp constant now short-circuit through
  `trivial_reduce` and pay a flat 50 JitCost (Scala's
  `EvalSigmaPropConstant`), instead of flowing through the generic
  constant-evaluation arm. This covers both the non-segregated form
  (`Expr::Const(SSigmaProp)`) and the segregated form
  (`Expr::ConstPlaceholder { resolved: Some(SigmaProp) }`), so every
  P2PK input pays the same cost regardless of how the script was
  serialized.
- **Placeholder and constant evaluation are priced separately.** A
  `ConstantPlaceholder` in a segregated tree is now evaluated in place
  at 1 JitCost (Scala's `ConstantPlaceholder = Fixed(1)`), not
  substituted into `Expr::Const` and charged 5 JitCost (Scala's
  `Constant = Fixed(5)`). The distinction exists in Scala's cost model
  so that segregation produces cheaper per-use references to repeated
  constants; sigma-rust previously erased the distinction at
  proposition time, overcharging every segregated script by 4 JitCost
  per constant reference. `ErgoTree::proposition_for_cost_eval()`
  preserves the placeholder nodes through reduction to restore this.
- **`BlockValue` charges `ADD_TO_ENV_COST` (5 JitCost) per `ValDef`
  insertion.** Each `let` binding pays the per-item base cost
  (`PerItemCost(1, 1, 10)` for the `BlockValue` itself) plus a flat 5
  JitCost for the environment extension, matching Scala's
  `AddToEnvironment` step. Prior to this, sigma-rust paid only the
  base and the rhs eval, silently undercounting every non-trivial
  script by `5 × n_val_defs`. Observed on mainnet tx
  `518acec…` @ height 700032: 4 ValDefs short-counted by 20 JitCost
  (2 block cost) before the fix; exact parity after. The same
  `ADD_TO_ENV_COST` constant is already charged per lambda invocation
  in Map/Filter/Fold/ForAll/Exists since Phase 6.
- **`Apply` charges `ADD_TO_ENV_COST` (5 JitCost) per lambda-arg
  binding.** `Apply::eval` binds each applied lambda argument into the
  interpreter env; each insertion now pays the flat 5 JitCost
  environment-extension cost, matching Scala's `FuncValue.eval` closure
  (`env1 = env + (arg -> vArg)`, `AddToEnvironmentDesc_CostKind =
  FixedCost(JitCost(5))`). Previously `Apply` paid only its `Fixed(30)`
  base and charged nothing for the bindings, undercounting every
  user-lambda application by `5 × n_args`: a single generic `Apply` ran
  5 JitCost short, a higher-order lambda chaining 5 applications ran 25
  short. Same `ADD_TO_ENV_COST` already charged for `BlockValue` ValDefs
  (above) and per-item coll-op lambdas (Phase 6); the coll-op path binds
  through its own closure and never routes through `Apply::eval`, so it
  is not double-charged.
- **`SigmaPropBytes` cost scales with the proposition's node count.**
  Now charges `PerItemCost(35, 6, 1)` over the SigmaBoolean node count
  (Scala's `numNodes = wrappedValue.size`), after the input eval, instead
  of a hardcoded `n = 1`. `SigmaBoolean::size`: ProveDlog/TrivialProp = 1,
  `ProveDHTuple` = 4 (one node per EcPoint), conjecture = `1 + children
  sizes`. The old `n = 1` undercharged every multi-node proposition.
- **`Coll.flatMap` cost is charged over the OUTPUT (flattened) length**,
  matching Scala's `flatMap_eval` (`addSeqCost` over `res.length` after
  building the result). sigma-rust previously charged over the input
  length before the map; flatMap output is typically ≫ input, so it
  undercounted.
- **`Coll.indexOf` cost reflects iterations performed**, not the full
  length. Scala's `indexOf_eval` charges `PerItemCost(20, 10, 2)` over
  `i - start` (from `max(from, 0)` to the found index inclusive, or the
  end). sigma-rust now scans first, then charges over `found - start + 1`
  (or `len - start` when not found).
- **Nested-collection equality recurses per element.** Scala's
  `DataValueComparer` bulk-compares COA leaf-element colls (`equalCOA_*`)
  but recurses `equalDataValues` per element for composite-element colls
  (Coll/Tuple/Option/SigmaProp — generic `equalColls`), charging the
  nested MatchType + per-item each time (short-circuiting on the first
  inequal element). `eq_with_cost` now recurses for the `coll_eq_cost`
  default arm (via `is_coa_coll`) and keeps the bulk compare for COA
  leaves — only the existing per-type constants, applied recursively.

## Phase-by-phase status

| Phase | Status | Commit |
|---|---|---|
| 1 — S4 per-tx accumulator | merged | `5e8cedcd` |
| 2 — S3 crypto cost | merged | `7ed2ba82` |
| 3 — S1+S2 init cost | merged | `40056345` |
| 4 — Bug 4+5 DataValueComparer | merged | `807e2c87` |
| 5 — Bug 3 SubstConstants count | merged | `285d8ba0` |
| 6 — Bug 6 lambda env cost | merged | `e1faa848` |
| 7 — Bug 7 Slice output size | merged | `5586f013` |
| 8 — Bug 2 trivial_reduce | merged | `0daa18a1` |
| 9a — Bug 1 IR plumbing (resolved field, proposition_for_cost_eval) | merged | `dc78df43` |
| 9b — Bug 1 interpreter switch + placeholder eval arm | merged | `233b66f9` |
| 10 — Test vectors / cost_parity harness | merged | pending commit |
| 11 — BlockValue ADD_TO_ENV_COST per ValDef | merged | pending commit |
| 12 — Apply ADD_TO_ENV_COST per lambda-arg binding | merged | `1e8d5845` |
| 13 — SigmaPropBytes cost scales with SigmaBoolean node count | merged | `c0c40ee6` |
| 14 — Coll.flatMap cost over output length | merged | `d404de55` |
| 15 — Coll.indexOf cost over iterations performed | merged | `6b48aff4` |
| 16 — nested-collection equality recurses per element | merged | `84676c07` |

mwaddip:
Superseded by #876 — *JIT Costing: The Definitive PR*, a clean 24-commit reslice of this work (same content, plus the folded SigmaProp equality conformance fix). Closing in favour of #876.

PR #871, opened 2026-06-01
fix: substConstants serializer-level parity: body-verbatim, OOB no-op, first-wins, v3 size-gate (match sigma-state)

mwaddip:
`SubstConstants` full-parsed the script bytes and errored when a position was >= the tree's constant count. sigma-state's `ErgoTreeSerializer.substituteConstants` is byte-surgery: it parses only the header + constants, keeps the body verbatim (never deserializing it), substitutes in-range positions and no-ops out-of-range ones. sigma-rust thus rejected scripts the JVM accepts — an OOB position or an unparseable body — a latent consensus fork (not seen on mainnet).

Adds `ErgoTree::substitute_constants` mirroring the Scala serializer; `SubstConstants::eval` calls it. Verified byte-exact vs JVM-blessed v5 vectors (5 OOB entries, incl. one with an unparseable body, return input unchanged); in-range covered by proptests. New `substitute_constants_oob_is_noop` test.

### DIFF
--- a/ergotree-interpreter/src/eval/subst_const.rs
+++ b/ergotree-interpreter/src/eval/subst_const.rs
@@ -59,9 +59,13 @@ impl Evaluable for SubstConstants {
             // parsed and out-of-range positions are a no-op, so a malformed
             // body or an OOB position returns the original bytes (JVM parity)
             // instead of erroring.
-            let (new_bytes, _num_constants) =
-                ErgoTree::substitute_constants(b.as_vec_u8(), &positions, &new_constants)
-                    .map_err(to_misc_err)?;
+            let (new_bytes, _num_constants) = ErgoTree::substitute_constants(
+                b.as_vec_u8(),
+                &positions,
+                &new_constants,
+                ctx.tree_version(),
+            )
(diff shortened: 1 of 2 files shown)


mwaddip:
Addendum: scope grew beyond the original out-of-range no-op to the full serializer-level substConstants rewrite (title updated). All four behaviors match sigma-state's ErgoTreeSerializer.substituteConstants. The v3 size-gate isn't mainnet-reachable, so it's covered by a unit test rather than a blessed vector.

mwaddip:
Added a third commit (65a98d12): substitute_constants parsed (and re-serialized) the template's constants under the template header's own version instead of the outer evaluation's tree version. The JVM ErgoTreeSerializer.substituteConstants reuses the single outer VersionContext (no inner re-entry per the template header), so a version-gated constant is accepted iff the OUTER tree allows it — sigma-rust over-accepted an outer-v2/template-v3 tree and rejected an outer-v3/template-v2 tree. Both directions now parse/re-serialize under the passed tree_version, pinned by JVM-blessed byte vectors.

PR #858, opened 2026-04-12
perf: lazy constant resolution in ErgoTree evaluation

mwaddip:
Resolve ConstPlaceholder from the eval context on demand instead of deep-cloning the AST via proposition() each reduce_to_crypto call. Mainnet-sync profiling in ergo-node-rust showed the per-input clone-and-substitute was 8-12% of total CPU.

Adds root_expr() / constants() accessors and Context::with_constants(). Deserialize trees fall back to proposition() (incompatible with lazy resolution). proposition() itself unchanged.

Beware of conflicts with #854, see comments below

PR #876, opened 2026-06-03
JIT Costing: The Definitive PR

### LINKED ISSUE #193: JIT costing
On evaluation every IR node should add their costs to `CostAccumulator`. Check out `direct-tree-eval` branch in sigmastate.

mwaddip:
Closes #193.

Ports JIT costing from sigmastate-interpreter (`direct-tree-eval`): each IR node
accumulates its Scala cost into the interpreter `Context` during evaluation, the
per-transaction cost limit is enforced cumulatively across inputs, and per-element
equality costs (`DataValueComparer`) and `SigmaByteWriter` serialize costs are
charged. A `cost_parity` test replays mainnet blocks 700000–700060 and asserts every
transaction's computed JIT cost equals the node's recorded cost (78 vectors).

**Supersedes #854.** Same logical change, re-sliced from #854's tangled history into
24 dependency-ordered, individually-green commits — each builds and passes the
`ergotree-interpreter` / `ergotree-ir` suites plus `cargo clippy --all-features
--all-targets -D warnings`. One conformance fix is folded in: `SigmaProp == SigmaProp`
now charges per Scala `equalSigmaBoolean` (ProveDlog 174 / ProveDHTuple 690) instead of
the flat catch-all 3.

### Relationship to #858 (lazy constant resolution)

Independent PR, overlapping in `reduce_to_crypto` / `ConstPlaceholder` handling — whichever
merges first, the other takes a mechanical rebase. This PR resolves segregated
`ConstPlaceholder` via `ConstantPlaceholder::resolved` (populated by
`ErgoTree::proposition_for_cost_eval`) and charges it `JitCost(1)`; #858 introduces
`Context.constants` lazy resolution. Reconciliation either way: the `ConstPlaceholder`
arm ends up charging **1 JitCost** and resolving via **`ctx.constants`**, dropping this
PR's `resolved`-field machinery as redundant. Worked reconciliation on the
`ergo-node-integration` branch (`9094f693`); full rebase recipe pinned on #858.


mwaddip:
Pushed 592d4f9c — degenerate-tree op-cost height. Scala's `CAvlTreeVerifier.treeHeight` reads `BatchAVLVerifier.rootNodeHeight`, assigned from the digest's trailing byte only after reconstruction's up-front requires (`keyLength > 0`, digest length) pass; when they fail, no root is built and the JVM charges a zero-height walk. We read the digest byte unconditionally, overcharging the height-scaled AvlTree ops on trees with non-positive keyLength (a signed Int on the JVM). Note: the degenerate path is reachable end-to-end only with ergoplatform/ergo_avltree_rust#14 (the crates.io verifier panics on these inputs before costing is reached), so in-repo coverage is a unit pin on tree_height; validated against JVM-blessed conformance vectors.

PR #880, opened 2026-06-04
fix(ergotree-interpreter): reduce atLeast with degenerate bound instead of erroring

mwaddip:
`atLeast(bound, children)` errored when `bound > children.size` or `bound < 0`. The JVM `AtLeast.reduce` (sigma/ast/trees.scala) reduces these to trivial props instead: `bound <= 0` => TrueProp, `bound > size` => FalseProp (any bound, incl. > 255 — the 255 cap is a `require` on the child count, not the bound). The guards short-circuited before `Cthreshold::reduce`, which already encodes this; the reorder just lets it run, with the u8 conversion only on the valid `1 <= bound <= size` path.

Found via a node syncing testnet: block 184,137 spends a V0 self-replicating contract evaluating `atLeast(1, Coll[SigmaProp]())` (empty conditions) — must be FalseProp, not an error. Adds 3 regression tests.

### DIFF
--- a/ergotree-interpreter/src/eval/atleast.rs
+++ b/ergotree-interpreter/src/eval/atleast.rs
@@ -41,16 +41,29 @@ impl Evaluable for Atleast {
             })
             .collect::<Result<Vec<SigmaBoolean>, TryExtractFromError>>()?;
 
-        let bound_u8: u8 = bound.try_into().map_err(|_| {
-            EvalError::Misc(format!("Atleast: bound is ({}) greater than 255", bound))
-        })?;
+        // Mirror Scala `AtLeast.reduce` (sigma/ast/trees.scala) degenerate-bound handling,
+        // applied before constructing a CTHRESHOLD:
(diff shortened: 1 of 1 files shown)


mwaddip:
Folded in a closely-related conformance fix on the same op: the 255-children cap.

`CSigmaDslBuilder.atLeast` (CSigmaDslBuilder.scala) throws `MaxChildrenCountForAtLeastOp = 255` as its first statement — *before* `AtLeast.reduce`, where the degenerate-bound arms this PR adds live. So the cap overrides the degenerate reductions: `atLeast(0, 256 props)` must error, not short-circuit to TrueProp. `ConcreteCollection` carries a u16 count, so >255 children are wire-constructible and reachable.

Added `input.len() > 255 ⇒ error` ahead of the bound arms, with a regression test (the 256-children rejects + the 255 cap-exclusive accept). The cap is exclusive (255 accepted); the bound itself stays uncapped.

PR #897, opened 2026-06-10
fix: reject unsupported (arity != 2) tuple-typed values at eval (JVM checkType)

mwaddip:
The JVM checkTypes each Tuple item (values.scala:801/804) via SType.isValueOfType, which rejects a tuple type of arity != 2 with "Unsupported tuple type" (SType.scala:200-205). sigma-rust models flat N-ary tuples (2..=255), so an arity-3 tuple carried as a constant (the item of a valid pair) evaluated where the JVM rejects — a consensus accept/reject divergence on crafted trees.

Adds check_value_type (eval.rs) mirroring isValueOfType's reachable rejections, called on Tuple-eval items. Constants are substituted into the body before eval on develop, so the segregated-constant case reduces to the same inline Tuple-item case.

Regression tests drive the JVM-blessed vector bytes end-to-end (parse → eval → expect the eval error).

### DIFF
--- a/ergotree-interpreter/src/eval/tuple.rs
+++ b/ergotree-interpreter/src/eval/tuple.rs
@@ -27,7 +27,18 @@ impl Evaluable for Tuple {
                 self.items.len()
             )));
         }
-        let items_v = self.items.try_mapped_ref(|i| i.eval(env, ctx));
+        let items_v = self
+            .items
+            .try_mapped_ref(|i| -> Result<Value<'ctx>, EvalError> {
+                let v = i.eval(env, ctx)?;
+                // Mirror the JVM Tuple eval, which `checkType`s each item
+                // (values.scala:801/804): an item whose type is an unsupported tuple
+                // (arity != 2) is rejected ("Unsupported tuple type", SType.scala:200).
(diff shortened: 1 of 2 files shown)


ergotree-interpreter/src/eval/tuple.rs
```diff
@@ -12,7 +12,18 @@ impl Evaluable for Tuple {
...
+        let items_v = self
+            .items
+            .try_mapped_ref(|i| -> Result<Value<'ctx>, EvalError> {
+                let v = i.eval(env, ctx)?;
+                // Mirror the JVM Tuple eval, which `checkType`s each item
+                // (values.scala:801/804): an item whose type is an unsupported tuple
+                // (arity != 2) is rejected ("Unsupported tuple type", SType.scala:200).
+                // Catches an arity-3 tuple constant carried as a pair item, which the
+                // JVM refuses but sigma-rust would otherwise evaluate.
+                crate::eval::check_value_type(&i.tpe())?;
```
kushti, line 39: Better to check type before evaluation to abort earlier if type is not ok?

kushti (changes requested):

kushti (approved):

kushti (approved):

